PR #31, opened 2019-05-20
Docs/readme system requirements

MatthewJMcAllister:
Edited section to improve consistency, and specified 64-bit requirement of processors based on user call.

Also testing pull request from Denver's fork to the Develop branch with USDOT :)


alexsoble:
@MatthewJMcAllister Taking another look at this — I think the reason we have some extra commits along for the ride here may be because we started editing README on `master` instead of starting from `develop`.

Perhaps the V2X Hub team will be able to rebase or cherry-pick your edit so that the pull request tells a cleaner story. Or we could open a fresh PR for a cleaner history. 

... also, hello V2X Hub team! 😄 👋

PR #75, opened 2019-10-01
Feature/pedestrian plugin

arayama:
The pedestrian plugin is able to register PSM with the webservice, call appropriate apis for broadcasting to the RSU.


snallamothu:
Dockerfile changes are using Ubuntu 18 image but pulling gcc5 instead of gcc7.  Please check if we have any others that needs to be corrected by comparing against the latest  Docketfile on develop,

PR #110, opened 2019-11-12
Feature/preemption plugin

arayama:
Changes include - 

BSM logger will not crash for any type of BSMs. ( short or extended) 
Preemption plugin first checks for extended packets than looks for sirens status.


src/v2i-hub/BsmLoggerPlugin/src/BsmLoggerPlugin.cpp
```diff
@@ -169,14 +169,6 @@ void BsmLoggerPlugin::HandleBasicSafetyMessage(BsmMessage &msg,
 	bool sirensOK = false;
 	bool lightsOK = false;
 
-	if(bsm->partII[0].list.array[0]->partII_Id >= 0)
```
snallamothu: Should this check be >0 or >=0?

PR #121, opened 2020-01-06
Feature/performance measures

jtbaird:
PR to merge Amir's changes in feature/PerformanceMeasures into develop.


arayama:
I think we need to avoid using absolute paths in the python codes ((e.g BSM_MAP.py). We can define a variable and keep it empty (" ") and use that variable for path reference to a file or folder. Any user can edit that variable as per their needs.

Issue #174, opened 2021-03-18
Update CircleCI with step to build and push docker image on commit 


paulbourelly999:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [ x ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary
Add step in CircleCI script to build and push docker image to DockerHub on github code changes. On failure, fail build to maintain git branches with buildable images and to allow testing without downloading source code. An additional step that may be worth investigating to allow testing branches without source code will be maintaining mysql images for each branch as well. This is not necessary for this issue but would be helpful and is worth investigating the effort required for this. Also worth investigating would be deploying images on CircleCI and performing some simple health checks. 

### V2XHUB version where this issue was discovered

5.0
### Expected behavior
Every commit will trigger CircleCI to build, tag and push docker image to V2XHub DockerHub. Any failure in this process will be considered a failed CircleCI build. This will allow us to keep track of the state of our build process and force us to maintain the Dockerfile in a working state.
### Actual behavior

Currently CircleCI only builds V2XHub locally and never actually runs the Dockerfile.

### Steps to reproduce the actual behavior

1. Commit code to remote
2. Triggers CircleCI script

PR #177, opened 2021-04-07
Docker Build CircleCI Script

paulbourelly999:
# PR Details
## Description
Added CircleCI script that builds and pushes v2xhub docker images with tags based on branch name.

## Related Issue

#174 
## Motivation and Context

Previously the CircleCI script only compiled the v2xhub code partially. The new script using the dockerfile to completely build and push the docker image for v2xhub which allows us to instantly diagnose docker build failures and allows for easier testing in tagged remote images for each git branch. 
## How Has This Been Tested?
Commit and ran CircleCI script and successfully pushed docker image to docker hub.

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -149,14 +149,14 @@ jobs:
           cd configuration/amd64/
           docker-compose up -d db
           cd ../../
-          docker build --network=host -t usdotfhwaops/v2xhubamd:$CIRCLE_BRANCH .
+          docker build --network=host -t usdotfhwaops/v2xhubamd:${CIRCLE_BRANCH,,} .
     
      - run: 
         name: Docker Push
         # Push v2xhub docker image to usdotfhwaops dockerhub
         command: | 
           echo "$DOCKERHUB_PASSWORD" | docker login --username $DOCKERHUB_USERNAME --password-stdin
-          docker push usdotfhwaops/v2xhubamd:$CIRCLE_BRANCH
+          docker push usdotfhwaops/v2xhubamd:${CIRCLE_BRANCH,,}
   docker_build&push_develop :
    machine:
     image: ubuntu-1604:202004-01
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,7 +58,6 @@ jobs:
             make
             make install
             cd ../../../..
-
             cd /home/V2X-Hub-build/ext
             git clone https://github.com/usdot-fhwa-OPS/qhttpengine.git
             cd /home/V2X-Hub-build/ext/qhttpengine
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -23,120 +23,29 @@ version: 2
...
-            set -x
-            source /home/V2X-Hub-build/.base-image/init-env.sh
-            cd /home/V2X-Hub-build/src
-            build-wrapper-linux-x86-64 --out-dir ../bw-output ./build.sh
-      - run:
-          name: Run Tests & Generate Gcovr Reports
-          command: |
-            set -x
-            cd /home/V2X-Hub-build/src
-            ./coverage.sh
```
paulbourelly999, line 119: Initially I thought everything I was removing from the CircleCI script was duplicated somewhere in the dockerfile but it looks like a couple things like this and the Run Sonar Scanner step may be unique to the CircleCI script so I will investigate their purpose and see if I need to keep them. I will also investigate the sonar-scanner checkout and build in the dockerfile to see if that is necessary since we don't run the sonar-scanner in the dockerfile

.circleci/config.yml
```diff
@@ -23,120 +23,29 @@ version: 2
...
-      - run:
-          name: Run Sonar Scanner
```
paulbourelly999, line 121: Might be necessary to keep this step since it is not replicated in the dockerfile. Also will investigate whether the V2XHub code needs to be compiled to be scanned by the sonar scanner tool.

.circleci/config.yml
paulbourelly999: Added the old script back as well and will write an issue to investigate Sonar-scanner tool separately for Dockerfile and CircleCI script.

.circleci/config.yml
paulbourelly999: Added the old script back as well and will write an issue to investigate Sonar-scanner tool separately for Dockerfile and CircleCI script

dan-du-car: @paulbourelly999  I do not think we need a docker push to  push the image to dockerhub at the circleCI config. The dockerhub itself will automatically pickup the changes from a configured github branch.

dan-du-car (approved):

jwillmartin (approved):

paulbourelly999:
Commits are squashed and branch is ready to be merged when the last reviewer is ready to merge it. Check with @arayama  and @animeshb27 before merge because I believe they are testing the develop branch after the CARMACloudPlugin merge and may want to hold off on any other merges into develop until their testing wraps up.

SaikrishnaBairamoni (approved):

SaikrishnaBairamoni:
Merged.

Issue #183, opened 2021-05-10
CARMACloud Plugin merge bugs


paulbourelly999:
### Types of Issue

- [ x ] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

After the merge of the CARMAcloudPlugin two bugs were discovered. The first is that the sudo command is no longer recognized in the docker build process due to removing it from the apt-get install list. The second is that the tmxctl --plugin-install commands fail due to missing password in mysql connection calls. This is due to the "GetPw()" method in DbConnectionPool.cpp being commented out. This was accidently left in the branch but warrants further investigation since there were issues mentioned relating to the tmx looping endlessly when incorrect password is provided.
### V2XHUB version where this issue was discovered

5.0
### Expected behavior
The V2X-Hub Dockerfile builds
### Actual behavior

The V2X-Hub Dockerfile fails to build, first due to missing a password for mysql and then due to not recognizing sudo as a command.
### Steps to reproduce the actual behavior

1. Checkout develop
2. Navigate to configuration/amd64/
3. Run docker-compose up -d db
4. Navigate to V2X-Hub/
5. Run docker build --network=host -t <image name>:<tag name> .
6. Build fails on tmxctl commands or on sudo commands.

PR #185, opened 2021-05-11
CARMACloud Plugin merge bugs

paulbourelly999:
# PR Details
## Description

These are the fixes for the CARMA cloud plugin merge to develop. An unintended change DbConnectionPool.cpp and ivpcore.cpp was that the contents of the GetPwd() method, which is meant to retrieve the MySQL password from an environment variable, was commented out. With these commented out the Dockerfile build fails during the tmxctl --install-plugin commands since these attempt to authenticate to MySQL and without a password this is not possible. The other fix included is the addition of sudo to the apt-get install list. Prior to the CARMA cloud plugin merge this was included, but it seems to be an oversight that this was left out. Leaving this out causes the Dockerfile build to fail when sudo commands are call since it does not recognize sudo command. Both these changes have been reverse to fix the Dockerfile build.
## Related Issue
#183 

## Motivation and Context

Fix the Dockerfile build
## How Has This Been Tested?
Building Dockerfile

## Types of changes

- [ X ] Defect fix (non-breaking change that fixes an issue)

## Checklist:

- [ X] I have read the **CONTRIBUTING** document.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM ubuntu:bionic-20190807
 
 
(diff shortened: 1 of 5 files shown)


arayama (approved):
I think the password issue will have to be solved using a different approach. The env for now makes sense.

PR #225, opened 2021-07-21
CF-143: Update Message schema, tables and logic

paulbourelly999:
CF-143 In progress

CF-143: save point

Save point fixed port_drayage.sql

CF-143: Edited database tables and added new action with destination

CF-143: Fixed and clean logic for new message schema

# PR Details
## Description

After redesigning the port_drayage MobilityOperation message payload and accounting for initial vehicle interactions the logic inside the MobilityOperation Plugin had to be update. The design changes are as follow. To account for initial vehicle and V2X-Hub interactions there now exists a first_action table. This table is used to take initial port_drayage messages coming from vehicles that are communication with the MobilityOperation plugin for the first time, that contain CMV id and look up the first action which will contain a linking next action. To discern first communication from normal communication , port_drayage messages will not included action_id value if they are initial communication with the MobilityOperation Plugin. Once the first action is returned to the vehicle, every subsequent response will include the complete action id, which V2X-Hub will use to find the next action from the freight table. This means that next_action is no longer part of the port_drayage json, and now just internal to V2X-Hub and the PORT_DRAYAGE DB for linking action. 

The PORT_DRAYAGE now contains two tables, the first_action table and the freight table. The first_action table contains all the initial actions for vehicles and allows the MobilityOperation plugin to lookup first actions based on cmv_id. The freight table contains all actions, including the first action for any vehicle and is used to look up the completed action and the next action using action_id and next_action. We include the first action again in the freight table to be able to link completion of the first action to its next action since, if the first action id is not present in the freight table we would not find it when a vehicle reports it has completed it.

Some changes to the data type of the json port_drayage payload include changing cmv_id to int, long and lat values to int, and action_id and next_action to string UUIDs.

## How Has This Been Tested?

This has been tested by using the following MobilityOperation messages. Sending these to the MessageReceiver Plugin and enabling the MobilityOperation Plugin and logging the received messages and the response
## Types of changes

## Checklist:


src/v2i-hub/MobilityOperationPlugin/src/MobilityOperationPlugin.cpp
```diff
@@ -233,12 +231,166 @@ void MobilityOperationPlugin::createMobilityOperationXml( ptree &mobilityOperati
...
+	
+}
+
+/**
+ * Retreive first action from first_action table using cmv_id.
+ * 
+ * @param cmv_id 
+ * @return PortDrayage_Object of first action 
+ */
+MobilityOperationPlugin::PortDrayage_Object MobilityOperationPlugin::retreiveFirstAction( uint32_t cmv_id ) {
```
JonSmet, line 306: Typo for the word 'retrieve'. Please fix here and in the other locations within the code included in this PR as well.

src/v2i-hub/MobilityOperationPlugin/src/MobilityOperationPlugin.h
```diff
@@ -51,23 +52,29 @@ class MobilityOperationPlugin: public PluginClient {
...
-	void createPortDrayageJson( PortDrayage_Object &pd_obj, ptree &json_payload);
+	ptree createPortDrayageJson( PortDrayage_Object &pd_obj);
 
-	void createMobilityOperationXml( ptree &mobilityOperationXml, ptree json_payload);
+	ptree createMobilityOperationXml( ptree &json_payload);
 
 	void HandleMobilityOperationMessage(tsm3Message &msg, routeable_message &routeableMsg);
 
-private:
+	PortDrayage_Object retreiveNextAction( std::string action_id );
```
JonSmet, line 61: I am not sure what the normal header file style is for V2XHub, but it could be good to include the function header comments from the .cpp file for each of the function declarations in the header just to provide someone with context when reading through the file.

JonSmet (changes requested):

PR #231, opened 2021-08-03
201 Remove Plain Text Passwords

paulbourelly999:
# PR Details
## Description

The currently deployment of V2X-Hub contains plain text password credentials in several available locations. It includes MySQL root password credentials in the Dockerfile and in the docker-compose.yml. Scripts like `V2X-Hub/data/DatabaseSetup/install_db.sh` and `V2X-Hub/configuration/arm64/mysql/install_db.sh` also include MySQL root credentials and V2X-Hub credentials. The initialization.sh script, ( located under /V2X-Hub/configuration/arm64 or  /V2X-Hub/configuration/amd64 ) does not include plain text information but instead writes this plain text into a .env file for use by the docker-compose.yml.
Besides the obvious security threat this creates, I believe it creates additional issues. Having our V2X-Hub docker images contain built-in MySQL root credentials makes configuring custom credentials more difficult since we would have to make modifications on the MySQL image and on the V2X-Hub image.

This propose solution removes many of these scripts, removes all creation of .env files since the value of set .env values can be found with a simple `docker inspect <container>'  call. Instead this solution uses docker secrets, which are essentially mounts, the same as the .env, except their content is not viewable through docker inspect. 

In addition, this solution removes the V2X-Hub image's dependency on MySQL. Initially, the V2X-Hub image was dependent on MySQL because as part of the image build, there were steps to install all the V2X-Hub plugins which requires access to a running MySQL database to save configurations and state. I have moved these calls to the service.sh script, where I believe they are more appropriate.

Finally, this solution also removes the script-credentials.sh script, which was added to the MySQL docker-entrypoint.d directory, where it was executed on the startup of the MySQL image to automatically configure an initial V2X-Hub user. The script did not work well with secrets, since the environment variable for any secret resolves to the location of the secret mount and not a value. In addition, this requires you to enter the root password in plain text on mysql, which would appear in the mysql container logs. As a replacement, there is a new add_v2xhub_user.sh script located under `V2X-Hub/configuration/amd64/mysql`. This script is run on the local machine and prompts the user to create a new V2X-Hub user in the MySQL database.

I have updated the deployment instructions to include the new deployment instruction on confluence here (https://usdot-carma.atlassian.net/wiki/spaces/V2XH/pages/1886158849/V2X-Hub+Docker+Deployment)

## Related Issue
#231 

## Motivation and Context

This change has been implemented to improve security and remove the V2X-Hub build process's dependency on a running MySQL server.
## How Has This Been Tested?
This has been tested by following these deployment instructions.

1. Clone V2X-Hub github (branch)
2. Navigate to `V2X-Hub/configuration/amd64/`
3. Create a secrets directory and create the following text files with the following credentials inside (SN: ensure there are no newline characters or spaces)
      -- mysql_root_password.txt
      -- mysql_password.txt

## Types of changes

## Checklist:

- [ x ] My change requires a change to the documentation.
- [ x ] I have updated the documentation accordingly.


.circleci/config.yml
```diff
@@ -161,10 +161,7 @@ jobs:
         name: Docker Build
         # Run MYSQL image and build v2xhub docker image
         command: |
-          cd configuration/amd64/
-          docker-compose up -d db
-          cd ../../
-          docker build --network=host -t usdotfhwaops/v2xhubamd:${CIRCLE_BRANCH,,} .
+          docker build -t usdotfhwaops/v2xhubamd:${CIRCLE_BRANCH,,} .
```
snallamothu, line 164: Is the docker-compose command no longer needed?  With us having have to handle ARM also now, I would think it's needed.

.circleci/config.yml
```diff
@@ -181,10 +178,7 @@ jobs:
         name: Docker Build
         # Run MYSQL image and build v2xhub docker image develop image
         command: |
-          cd configuration/amd64/
-          docker-compose up -d db
-          cd ../../
-          docker build --network=host -t usdotfhwaops/v2xhubamd:latest .
+          docker build -t usdotfhwaops/v2xhubamd:latest .
```
snallamothu, line 181: Same comment as above. Please confirm docker-compose no longer need to be run.

src/tmx/TmxUtils/src/database/DbConnectionPool.cpp
```diff
@@ -120,17 +120,23 @@ void DbConnectionPool::SetConnectionUrl(std::string connectionUrl) {
 }
 
 std::string DbConnectionPool::GetPwd(){
-	const char* EnvVar = "MYSQL_ROOT_PASSWORD";
+	const char* EnvVar = "MYSQL_PASSWORD";
```
snallamothu, line 123: This is an improvement over what we have currently.  Are we going to revisit the strategy of using environment variables for passwords when we implement security because IMO this is still a risk because anybody that gains access to the operating system can query for it?

snallamothu (changes requested):

.circleci/config.yml
paulbourelly999: No I have removed the dependency on mysql during the build process since we have moved the plugin install calls into the run script. I believe this is a better way to do this as well since it allows the build to happen independently or mysql and there are similar calls in the run script anyway.

.circleci/config.yml
paulbourelly999: It is no longer needed. It was needed before since we were installing plugins/ adding configurations to mysql during the build process. We know only do that on startup. This allows the build image not to need or be bound to mysql credentials.

src/tmx/TmxUtils/src/database/DbConnectionPool.cpp
paulbourelly999: Slight change is that now the environment variable points to a file which is mounted to the container. The actual file can be located on an HSM, the environment variable is only necessary to pass the file location to the container.

snallamothu (approved):

Issue #228, opened 2021-07-30
Create custom PHP image without volumes


paulbourelly999:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [ x ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

Currently, to deploy V2X-Hub's web UI we use a PHP image and pass it our HTML and JS using a volume in the docker-compose. This means our PHP image relies on a local web folder to be present to be deployable. By creating a dockerfile for a custom PHP image we can add this web directory to the image directly and avoid reliance on a local web directory for deployment.

### V2XHUB version where this issue was discovered

6.0
### Expected behavior
V2X-Hub PHP image is independently deployable without reliance on a local web directory
### Actual behavior

We deploy the stock PHP image and provide our web code via a docker-compose volume.
### Steps to reproduce the actual behavior

1. Remove your local /V2X-Hub/web/ directory
2. Attempt to deploy V2X-Hub using docker-compose
3. PHP image will fail and no web UI will be deployed

PR #232, opened 2021-08-03
Issue-228:No volume PHP image

paulbourelly999:
Issue-228:Corrected Docker-compose

# PR Details
## Description

Added CirclCI and Dockerfile to create a custom php image to alleviate the dependency on a local web/ directory for the deployment of the PHP image for V2X-Hub. Before hand, we used the stock PHP image and loaded our own JS and HTML in via a large `web/` volume, which meant that in order to deploy our image we need the local web directory. This dependency is unnecessary when building a custom PHP image and became an issue when configuring new docker-compose files for services like carma-streets which require V2X-Hub. 
## Related Issue

#228 
## Motivation and Context

Added CirclCI and Dockerfile to create a custom php image to alleviate the dependency on a local web/ directory for the deployment of the PHP image for V2X-Hub. Before hand, we used the stock PHP image and loaded our own JS and HTML in via a large `web/` volume, which meant that in order to deploy our image we need the local web directory. This dependency is unnecessary when building a custom PHP image and became an issue when configuring new docker-compose files for services like carma-streets which require V2X-Hub.
## How Has This Been Tested?

This was tested by deploying the new PHP image and testing it's functionality locally. This can be done by changing the image and tag name inside the docker-compose file to usdotfhwaops/php:228-no-volume-php .
## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -164,6 +164,7 @@ jobs:
           cd configuration/amd64/
           docker-compose up -d db
           cd ../../
+          docker build -f docker/php.yml -t usdotfhwaops/php:${CIRCLE_BRANCH,,} .
           docker build --network=host -t usdotfhwaops/v2xhubamd:${CIRCLE_BRANCH,,} .
     
      - run: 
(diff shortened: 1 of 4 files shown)


Dockerfile-php-no-volume.yml
```diff
@@ -0,0 +1,3 @@
+FROM php:7.2.2-apache
+COPY web/ /var/www/html/
+WORKDIR /usr/local/bin/
```
dan-du-car, line 3: If WORKDIR command is not needed, it shall be removed.

PR #240, opened 2021-08-31
Rename to PortDrayagePlugin

paulbourelly999:
+ Fix segfault with not found actions
+ Fix CMakefile
+ Fix sql file to only include necessary information

# PR Details
## Description

Renamed CARMAFreightPlugin to PortDrayagePlugin. Fixed Segmentation Fault on empty responses from the PORT_DRAYAGE DB
## Types of changes

## Checklist:


src/v2i-hub/PortDrayagePlugin/src/PortDrayagePlugin.cpp
```diff
@@ -223,27 +227,28 @@ CARMAFreightPlugin::PortDrayage_Object CARMAFreightPlugin::retrieveNextAction( s
 			sql::ResultSet *cur_action = next_action->executeQuery();
 			// Create PortDrayage_Object
 			PortDrayage_Object *rtn = new PortDrayage_Object();
-			if ( cur_action->first() )  {
```
JonSmet, line 226: This check was deleted, but shouldn't we still be checking whether an sql::ResultSet exists for the next action id?

src/v2i-hub/PortDrayagePlugin/src/PortDrayagePlugin.cpp
```diff
@@ -209,7 +213,7 @@ ptree CARMAFreightPlugin::createMobilityOperationXml( ptree &json_payload ) {
 }
 
 
-CARMAFreightPlugin::PortDrayage_Object CARMAFreightPlugin::retrieveNextAction( std::string action_id ) {
+PortDrayagePlugin::PortDrayage_Object PortDrayagePlugin::retrieveNextAction( std::string action_id ) {
 	try{
 		// Set action_id
 		next_action_id->setString(1,action_id);
```
JonSmet, line 219: Since the 'action_id' passed to this function is the previously completed action id by the CMV, maybe the name for this object could be previous_action_id or current_action_id rather than next_action_id?

src/v2i-hub/PortDrayagePlugin/src/PortDrayagePlugin.cpp
```diff
@@ -223,27 +227,28 @@ CARMAFreightPlugin::PortDrayage_Object CARMAFreightPlugin::retrieveNextAction( s
 			sql::ResultSet *cur_action = next_action->executeQuery();
```
JonSmet, line 227: I know this is code from a previous PR, but it might be helpful to rename this to 'next_action' or something similar, if you agree.

src/v2i-hub/PortDrayagePlugin/src/PortDrayagePlugin.cpp
JonSmet: Additionally, if the 'next_action_id' object is renamed here, then the 'next_action' object on line 225 could be renamed to 'next_action_id' to maintain consistency.

src/v2i-hub/PortDrayagePlugin/src/PortDrayagePlugin.cpp
paulbourelly999: So this variable is a class field that represent a query to retrieve and action by action ID. I have changed it to be more appropriately named

PR #261, opened 2021-10-06
CARMA streets plugin to accept MobilityPath message

dan-du-car:
# PR Details
## Description
Add functionality in CARMA streets plugin to accept MobilityPath message. 
Fix the J2735 MobilityPath definition.
Construct and publish MobilityPath message upon receiving the message from MessageReceiver.

## Related Issue
https://github.com/usdot-fhwa-OPS/V2X-Hub/issues/262

## Motivation and Context
Cooperative stop use case

## How Has This Been Tested?
Local integration testing
Tested in white Ford Fusion

## Types of changes

## Checklist:

### DIFF
--- a/src/v2i-hub/CARMAStreetsPlugin/CMakeLists.txt
+++ b/src/v2i-hub/CARMAStreetsPlugin/CMakeLists.txt
@@ -2,4 +2,4 @@ PROJECT ( CARMAStreetsPlugin VERSION 5.0 LANGUAGES CXX )
 
 BuildTmxPlugin ( )
 
-TARGET_LINK_LIBRARIES (${PROJECT_NAME} tmxutils rdkafka++)
+TARGET_LINK_LIBRARIES (${PROJECT_NAME} tmxutils  rdkafka++ jsoncpp)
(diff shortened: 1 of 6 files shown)


src/v2i-hub/CARMAStreetsPlugin/CMakeLists.txt
```diff
@@ -2,4 +2,4 @@ PROJECT ( CARMAStreetsPlugin VERSION 5.0 LANGUAGES CXX )
 
 BuildTmxPlugin ( )
 
-TARGET_LINK_LIBRARIES (${PROJECT_NAME} tmxutils rdkafka++)
\ No newline at end of file
+TARGET_LINK_LIBRARIES (${PROJECT_NAME} tmxutils  rdkafka++ jsoncpp)
```
paulbourelly999, line 5: @dan-du-car the only issue with this PR to me would be this library. I removed it from the docker/Dockerfile-depl which is the lightweight image we are trying to create since it is included in our dockerfile but used no where else. If we want to include this library for the CARMA-Streets plugin, which I am fine with, we would need to update the docker/Dockerfile-depl for ARM builds to also include the library. The reason this was not caught by circleCI is because your branch is behind develop. If you rebase on develop and push your changes you will see that the ARM build will fail without adding this to the docker/Dockerfile-depl

PR #343, opened 2022-03-10
Sample BSM plugin

dan-du-car:
# PR Details
Provide example code for how to use V2X-Hub Plugin Client's BroadcastMessage method to send a BSM.

## Description
Provide example code for setting values for the BSM and then constructing the correct object to be passed to the BroadcastMessage method.

## Related Issue
NA

## Motivation and Context
voices support

## How Has This Been Tested?
local integration testing

## Types of changes

## Checklist:


src/v2i-hub/SampleBSMPlugin/manifest.json
```diff
@@ -0,0 +1,22 @@
+{
+	"name":"SampleBSMPlugin",
+	"description":"Sample BSMplugin for making websocket connection with Sample BSM.",
+	"version":"3.0.0",
```
paulbourelly999, line 4: probably should update this to reflect V2X-Hub version or something but not a priority

paulbourelly999 (changes requested):
Everything looks good but this plugin probably does not need to merged into develop as an actual plugin since it is only used for reference. We could include under `examples/tmx-exampleapps/` if we want to include an example plugin for BSM creation/encoding and decoding

paulbourelly999 (approved):
Looks good.

Issue #351, opened 2022-04-13
PHP UI missing Configuration parameters


paulbourelly999:
### Types of Issue

- [ x ] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

Even with updated configuration parameters inside a manifest JSON, occasionally the PHP UI will reflect and old set of configuration parameter. I believe this is due to the configuration parameter values stored in the localhost.sql files 
### Expected behavior
UI is displaying up to date configuration values from the manifest.json of the plugin.
### Actual behavior

UI is displaying old configuration values from an old version of the plugin

PR #355, opened 2022-04-15
Issue-351:Remove plugin state from localhost.sql files

paulbourelly999:
# PR Details
## Description

Remove plugin state from localhost.sql configuration files to avoid localhost.sql plugin state or configuration values from replacing tmx plugin installed state or configuration values. When PHP starts up it checks the mysql database for configuration values and so keeping old configuration values in the table can sometimes cause the PHP to pickup old configuration parameters instead of newly installed ones. This fix should remove that possibility.
## Related Issue
#351 

## Types of changes

- [ x] Defect fix (non-breaking change that fixes an issue)

## Checklist:


dan-du-car:
I think the PR looks good if it works as the PR description states. Is the old parameters inserted from localhost.sql SQL. Where in code insert the newly installed parameters. Which PHP we are referring to? PHP image for the v2xhub web? or PHP script?

configuration/amd64/mysql/localhost.sql
```diff
@@ -78,7 +78,6 @@ CREATE TABLE `installedPlugin` (
 
 LOCK TABLES `installedPlugin` WRITE;
 /*!40000 ALTER TABLE `installedPlugin` DISABLE KEYS */;
```
dan-du-car, line 80: Should we remove the lock and unlock statements if we are not writing data into the tables?

paulbourelly999: We can remove the lock statements although functionally it would not impact anything. I see these sql files as possible avenues to restoring a specific state of V2X-Hub. Then then these lock statements offer an outline to how to create a state. But since these files are generated normally by mysqldump, and the lock statements will be generated as well I can remove then since they do not serve a purpose.

paulbourelly999:
> ks as the PR description states. Is the old parameters inserted from localhost.sql SQL. Where in code insert the newly installed parameters. Which PHP we are referring to? PHP image fo

I think what would happen is the php script which queries the sql database would get configurations before the install overwrites and the UI would then reflect an invalid state where configuration parameters were missing or old removed parameters were still present but I haven't  confirm this is the actual behavior. I have tested with these sql files and they definitely work.

dan-du-car (approved):

PR #360, opened 2022-04-19
Hotfix remove tcms for  duplicated tcr ids

dan-du-car:
# PR Details
## Description
Not broadcasting messages after you see another TCR with the same request id.
Add configuration parameter to restrict the number of TCMs broadcast within a time out period.

## Motivation and Context
freight

## How Has This Been Tested?
NA

## Types of changes

## Checklist:

### DIFF
--- a/src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
+++ b/src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
@@ -32,6 +32,7 @@ CARMACloudPlugin::CARMACloudPlugin(string name) :PluginClient(name) {
 	base_ack = "/carmacloud/tcmack";
 	method = "POST";
 	_not_ACK_TCMs = std::make_shared<multimap<string, tsm5EncodedMessage>>();
+	_tcm_broadcast_times = std::make_shared<std::map<string, int>>();
 	std::thread Broadcast_t(&CARMACloudPlugin::Broadcast_TCMs, this);
 	Broadcast_t.detach();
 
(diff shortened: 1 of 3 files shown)


paulbourelly999 (approved):
Looks good to me. @dan-du-car Please let me know once this has been tested and is ready for merge.

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
```diff
@@ -317,6 +331,16 @@ void CARMACloudPlugin::Broadcast_TCMs()
 			cur_time = 0;	
 			is_started_broadcasting = false;
 		}
+
```
snallamothu, line 335: Why can't we just update the counter for each req after broadcasting instead of in a separate loop?

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
```diff
@@ -293,7 +296,18 @@ void CARMACloudPlugin::Broadcast_TCMs()
 					PLOG(logINFO) << "Sent No ACK as Time Out: "<< sss.str() <<endl;
 					CloudSend(sss.str(),url, base_ack, method);					
 					break;
```
snallamothu, line 299: Should this be continue instead of a break?  You want to continue to the next iteration, correct?

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
```diff
@@ -262,7 +262,8 @@ void CARMACloudPlugin::Broadcast_TCMs()
 			std::lock_guard<mutex> lock(_not_ACK_TCMs_mutex);
 			for( auto itr = _not_ACK_TCMs->begin(); itr!=_not_ACK_TCMs->end(); ++itr )
 			{
-				string tcmv01_req_id_hex = itr->first;
+				string tcmv01_req_id_hex = itr->first;	
+				//Start recording time out period
 				if(!is_started_broadcasting)
```
snallamothu, line 268: I don't understand the start_time logic, shouldn't start_time be different for each message because that should be set to the first time the message was added to the queue or broadcast, is_started_broadcasting is set to true after the first message is processed and will stay true for the rest of the collection.  So, all messages will have the same start time and different curr times.

dan-du-car:
@snallamothu I did not see these comments before. But I am closing this PR due to another replacement PR: https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/362

PR #362, opened 2022-04-21
Hotfix Repeatedly broadcast TCM and timeout bugs

dan-du-car:
# PR Details
## Description
The TCMs should be broadcast upon receiving a response from carma-cloud. It should be handled inside the HandleResponse method.
Remove TCMs with the same request ids as TCR if receiving a duplicated TCR request.
Fix time out logic that cause segmentation fault.
Add config parameter for the thread sleep for.
Add configuration parameter and logic to control the number of TCMs being broadcast.

## Related Issue
https://github.com/usdot-fhwa-OPS/V2X-Hub/issues/364

## Motivation and Context
Freight testing

## Types of changes

## Checklist:

### DIFF
--- a/src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
+++ b/src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
@@ -363,6 +363,12 @@ void CARMACloudPlugin::BroadcastTCM(tsm5EncodedMessage& tsm5ENC) {
 
 bool CARMACloudPlugin::IsSkipBroadcastCurTCM(const string & tcmv01_req_id_hex, const string & tcm_hex_payload ) const
 {
+	//Skip repeatedly broadcasting  
+	if(_TCMRepeatedlyBroadCastTotalTimes == 0)
+	{
+			return true;
+	}
+
 	bool is_skip_cur_tcm = false;
 	bool is_tcm_hex_found = false;
 	auto tcms_metadatas = _tcm_broadcast_times->equal_range(tcmv01_req_id_hex);
--- a/src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
+++ b/src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
@@ -32,6 +32,7 @@ CARMACloudPlugin::CARMACloudPlugin(string name) :PluginClient(name) {
 	base_ack = "/carmacloud/tcmack";
 	method = "POST";
 	_not_ACK_TCMs = std::make_shared<multimap<string, tsm5EncodedMessage>>();
+	_tcm_broadcast_times = std::make_shared<std::map<string, int>>();
 	std::thread Broadcast_t(&CARMACloudPlugin::Broadcast_TCMs, this);
 	Broadcast_t.detach();
 
--- a/src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
+++ b/src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
@@ -303,8 +303,10 @@ void CARMACloudPlugin::Broadcast_TCMs()
 				if (_tcm_broadcast_times->count(tcmv01_req_id_hex) == 0)
 				{
 					_tcm_broadcast_times->insert({tcmv01_req_id_hex, 0});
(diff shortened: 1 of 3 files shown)


src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
```diff
@@ -229,6 +229,19 @@ void CARMACloudPlugin::CARMAResponseHandler(QHttpEngine::Socket *socket)
 	container.load<XML>(ss);
 	tsm5message.set_contents(container.get_storage().get_tree());
 	tsm5ENC.encode_j2735_message(tsm5message);
+	string enc = tsm5ENC.get_encoding();
```
snallamothu, line 232: This code is duplicated from the other Broadcast_TCMs method, can you pull it out into a separate method and call it from both places to avoid duplication?

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
```diff
@@ -299,36 +309,32 @@ void CARMACloudPlugin::Broadcast_TCMs()
 				msg->refresh_timestamp();
```
snallamothu, line 309: Consider revising the method name because it's doing more than broadcasting the TCMs, perhaps something more like TCMAckCheckAndRebroadcastTCM?

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
```diff
@@ -299,36 +309,32 @@ void CARMACloudPlugin::Broadcast_TCMs()
...
+				sss << "<?xml version=\"1.0\" encoding=\"UTF-8\"?><TrafficControlAcknowledgement><reqid> " << tcmv01_req_id_hex
+						<< "</reqid><msgnum></msgnum><cmvid></cmvid><acknowledgement>" << acknowledgement_status::acknowledgement_status__not_acknowledged
+						<< "</acknowledgement><description>" << _TCMNOAcknowledgementDescription
+						<< "</description></TrafficControlAcknowledgement>"; 
+				PLOG(logINFO) << "Sent No ACK as Time Out: "<< sss.str() <<endl;
+				CloudSend(sss.str(),url, base_ack, method);		
+
+				_not_ACK_TCMs->erase(tcmv01_req_id_hex);
+				//If time out, stop tracking the starting time of the TCMs being broadcast so far
+				_tcm_broadcast_starting_time->erase(tcmv01_req_id_hex);
```
snallamothu, line 335: Is this deleting all TCMs associated with this TCR or just the particular TCM that this applies to?

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
```diff
@@ -264,24 +263,20 @@ void CARMACloudPlugin::Broadcast_TCMs()
...
 				string tcmv01_req_id_hex = itr->first;	
-				//Start recording time out period
-				if(!is_started_broadcasting)
+				auto cur_time = (std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch())).count();				
+				if (_tcm_broadcast_starting_time->count(tcmv01_req_id_hex) == 0)
 				{
-					start_time = (std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch())).count();
-				}else{
-					cur_time = (std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch())).count();
+					_tcm_broadcast_starting_time->insert({tcmv01_req_id_hex, cur_time});
```
snallamothu, line 285: There can be multiple TCMs associated with a single TCM req ID and this is going to expire all of them which is probably fine in practice because they all will expire within msec apart but in theory, you should most likely use the TCM ID here instead as the key.  This probably applies to all checks.

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/363

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
dan-du-car: The time out should be for each TCR, right? After the vehicle sends a TCR and it receives the first TCM, we should count timer until it receives an ACK or time out.

Issue #374, opened 2022-05-18
Pedestrian plugin not sending multiple PSMs for multiple tracked pedestrians


abey-yoseph:
### Types of Issue

- [x ] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary
When there are multiple pedestrians being tracked in the crosswalk by the FLIR camera, V2X Hub is not sending all of the pedestrian info to the RSU (in the form of PSMs). It only creates a PSM for the last tracked pedestrian.

### V2XHUB version where this issue was discovered
Release_7.3

### Expected behavior
One PSM for each tracked pedestrian. 

### Actual behavior
Only one PSM no matter how many pedestrians present.

### Steps to reproduce the actual behavior
1. Multiple people walk through one of the west intersection FLIR camera detection zones
2. Log the RSU transmit interface and decode the PSMs to see only one unique pedestrian (alternatively can look at the v2xhub docker container logs and see that only one UPER encoded PSM is being sent to the RSU)

PR #376, opened 2022-05-18
fix for multiple pedestrian handling

abey-yoseph:
# PR Details
## Description

Now using queue to store all created PSMs for each pedestrian reported by FLIR camera. Pedestrian plugin reads from the queue, sends each PSM to the RSU, and then pops from the queue.

## Related Issue
#374 

## Motivation and Context
Need all pedestrian data.

## How Has This Been Tested?
Two pedestrians walked through the detection zone and I confirmed that multiple unique PSMs were being created and sent to the RSU for broadcast.

## Types of changes
- [x ] Defect fix (non-breaking change that fixes an issue)

## Checklist:

### DIFF
--- a/src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
+++ b/src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
@@ -276,6 +276,7 @@ namespace PedestrianPlugin
 
 		                std::lock_guard<mutex> lock(_psmLock);
                         psmxml = psm_xml_str;
+                        psmQueue.push(psmxml);
 
                         PLOG(logDEBUG) << "Sending PSM xml to BroadcastPsm: " << psmxml.c_str() <<endl;
 
@@ -317,10 +318,10 @@ namespace PedestrianPlugin
         std::cout << beast::make_printable(buffer_.data()) << std::endl;
     }
 
-    //returns the variable containing the psm xml 
-    string FLIRWebSockAsyncClnSession::getPSMXML() const
+    std::queue<std::string>& FLIRWebSockAsyncClnSession::getPSMQueue()
     {
-        return psmxml;
+        std::lock_guard<mutex> lock(_psmLock);
+        return psmQueue;
     }
 
     vector<int> FLIRWebSockAsyncClnSession::timeStringParser(string dateTimeStr) const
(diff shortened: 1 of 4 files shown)


src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
```diff
@@ -317,10 +318,10 @@ namespace PedestrianPlugin
         std::cout << beast::make_printable(buffer_.data()) << std::endl;
     }
 
-    //returns the variable containing the psm xml 
-    string FLIRWebSockAsyncClnSession::getPSMXML() const
+    std::queue<std::string>& FLIRWebSockAsyncClnSession::getPSMQueue()
     {
-        return psmxml;
+        std::lock_guard<mutex> lock(_psmLock);
```
snallamothu, line 323: You don't need a lock here because you are only returning the reference to the queue.  Doesn't hurt but it's not needed.

src/v2i-hub/PedestrianPlugin/src/PedestrianPlugin.cpp
```diff
@@ -132,16 +132,15 @@ int PedestrianPlugin::checkXML()
...
-				PLOG(logINFO) << "CheckXML: Broadcasting new CP PSM!" << std::endl;
-
-				BroadcastPsm(const_cast<char*>(currentXML.c_str()));
-				lastGeneratedXML = currentXML;
-			}
+		{	
+			//retrieve the PSM queue and send each one to be broadcast, then pop		
+			std::queue<string> &currentPSMQueue = flirSession->getPSMQueue();
+
+			for (int i = 0; i < currentPSMQueue.size(); i++)
```
snallamothu, line 139: You will need to lock the mutex here because the one in get method is out of scope and your lock would have been released.

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
```diff
@@ -276,6 +276,7 @@ namespace PedestrianPlugin
 
 		                std::lock_guard<mutex> lock(_psmLock);
                         psmxml = psm_xml_str;
+                        psmQueue.push(psmxml);
```
snallamothu, line 279: I think you should add a mutex lock here also if you make the change I suggested below.

abey-yoseph: An additional lock to the one two lines above?

snallamothu: No, I missed that you had one already.

src/v2i-hub/PedestrianPlugin/src/PedestrianPlugin.cpp
snallamothu: I am actually going to merge this now because on second thought I don't think you need may not need locks.  Lets test and see.

snallamothu (approved):

Issue #383, opened 2022-05-31
FLIR PSMs Dsecond field incorrectly being populated


abey-yoseph:
### Types of Issue
- [x ] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary
There are two places in the created PSM that use the "DSecond" J2735 field. This number is not being properly incremented (up to 60999) causing the UTC timestamp to be incorrect.

### V2XHUB version where this issue was discovered
Release 7.3

### Expected behavior
DSecond will increase until 60999 and then start back at 0.

### Actual behavior
It only increases to 1000.

### Steps to reproduce the actual behavior
Turn the Pedestrian Plugin on and walk through one of the detection zones. In the docker logs, you will see all of the produced PSMs for that test and the times associated with each data point. The "secMark" and "second" fields will never be larger than 1000.

PR #384, opened 2022-06-01
fix second field in generated PSM

abey-yoseph:
# PR Details
## Description
Create a new variable to calculate the total milliseconds in the current minute.

## Related Issue
#383 

## Motivation and Context
Secmark field in PSM needs to be corrected.

## How Has This Been Tested?
Collected data at the West Intersection and verified the seconds field in the PSM is properly being updated via the docker log.

## Types of changes

- [x ] Defect fix (non-breaking change that fixes an issue)

## Checklist:

- [x ] I have read the **CONTRIBUTING** document.


paulbourelly999:
Your change looks fine. There are 3 code smells that may be nice to fix now. The one about using empty instead of size can replace the size check in your if statement and your for loop i think. The other one is about const cast. Is there a reason why you need to cast this as a const. Here are two examples of casting a string to a char* I found only. Would either work for you?

```
char* char_arr;
string str_obj("GeeksForGeeks");
char_arr = &str_obj[0]
```

```
// assigning value to string s
string st = "GeeksForGeeks";
//the c_str() function returns a const pointer
//to null terminated contents.
const char *str = st.c_str();
```

paulbourelly999 (approved):
Looks good

PR #387, opened 2022-06-14
Carma streets Plugin publishes map msp

dan-du-car:
# PR Details
Update CARMA-Streets V2XHub Plugin to forward MAP msg to CARMA-Streets

Convert MAP xml to JSON

Use the same frequency as V2X-Hub MAPPlugin
## Related Issue
NA

## Motivation and Context
UC 3

## How Has This Been Tested?
Local integration testing to verify that the MAP JSON format message is sent to a kafka topic and a console consume can print out the JSON message. Verified using intersection model to make sure the signal group ids are updated.
image

## Types of changes
Add feature to broadcast MAP message to kafka topic

## Checklist:


src/v2i-hub/CARMAStreetsPlugin/src/J2735MapToJsonConverter.cpp
```diff
@@ -0,0 +1,267 @@
...
+        const IntersectionGeometryList *intersections = mapMsgPtr->intersections;
+        Json::Value intersectionsJson;
+        // Assume there is only one intersection geometry for each intersection
+        for (size_t i = 0; i < intersections->list.count; i++)
+        {
+            Json::Value intersectionJson;
+            if (intersections->list.array != nullptr)
+            {
+                auto intersection = intersections->list.array[i];
+                intersectionJson["id"]["id"] = std::to_string(intersection->id.id);
```
paulbourelly999, line 32: What element does id id reference. Does the MAP message have an id and then include intersection id's as well. Is there a sample XML you could send me or include in some documentation as reference to what the JSON structure of outgoing map messages will be.

src/v2i-hub/CARMAStreetsPlugin/src/J2735MapToJsonConverter.cpp
```diff
@@ -0,0 +1,267 @@
...
+        // Construct Map Data
+        Json::Value mapDataJson;
+        mapDataJson["layer_id"] = mapMsgPtr->layerID;
+        mapDataJson["msg_issue_revision"] = std::to_string(mapMsgPtr->msgIssueRevision);
+        mapDataJson["layer_type"] = std::to_string(*mapMsgPtr->layerType);
+
+        // Construct intersections
+        const IntersectionGeometryList *intersections = mapMsgPtr->intersections;
+        Json::Value intersectionsJson;
+        // Assume there is only one intersection geometry for each intersection
```
paulbourelly999, line 25: Is this comment accurate. It looks like the for loop considers multiple intersections.

src/v2i-hub/CARMAStreetsPlugin/src/J2735MapToJsonConverter.cpp
```diff
@@ -0,0 +1,267 @@
...
+        Json::Value metadata;
+        auto timestamp_utc = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
+        metadata["timestamp"] = std::to_string(timestamp_utc);
+        mapJson["metadata"] = metadata;
+
+        // Construct Map Data
+        Json::Value mapDataJson;
+        mapDataJson["layer_id"] = mapMsgPtr->layerID;
+        mapDataJson["msg_issue_revision"] = std::to_string(mapMsgPtr->msgIssueRevision);
+        mapDataJson["layer_type"] = std::to_string(*mapMsgPtr->layerType);
```
paulbourelly999, line 20: Why does the mapMsgPtr need to be dereferenced here and not in the line above?

src/v2i-hub/CARMAStreetsPlugin/src/J2735MapToJsonConverter.cpp
```diff
@@ -0,0 +1,267 @@
...
+        }
+        mapDataJson["intersections"] = intersectionsJson;
+        mapJson["map_data"] = mapDataJson;
+    }
+
+    void J2735MapToJsonConverter::convertLanesetToJSON(const IntersectionGeometry *intersection, Json::Value &laneSetJson) const
+    {
+        // Construct laneset
+        const auto laneSet = intersection->laneSet;
+        if (laneSet.list.array != nullptr)
```
paulbourelly999, line 54: Does array not have an empty method?

src/v2i-hub/CARMAStreetsPlugin/src/J2735MapToJsonConverter.cpp
```diff
@@ -0,0 +1,267 @@
...
+        mapDataJson["layer_type"] = std::to_string(*mapMsgPtr->layerType);
+
+        // Construct intersections
+        const IntersectionGeometryList *intersections = mapMsgPtr->intersections;
+        Json::Value intersectionsJson;
+        // Assume there is only one intersection geometry for each intersection
+        for (size_t i = 0; i < intersections->list.count; i++)
+        {
+            Json::Value intersectionJson;
+            if (intersections->list.array != nullptr)
```
paulbourelly999, line 29: Does this array object not have an empty() method

src/v2i-hub/CARMAStreetsPlugin/src/J2735MapToJsonConverter.cpp
```diff
@@ -0,0 +1,267 @@
...
+            for (size_t i = 0; i < laneSet.list.count; i++)
+            {
+                std::stringstream ss;
+                Json::Value laneJson;
+                auto lane = laneSet.list.array[i];
+                laneJson["lane_id"] = std::to_string(lane->laneID);
+                if (lane->ingressApproach != nullptr)
+                {
+                    ss.str("");
+                    ss << *lane->ingressApproach;
```
paulbourelly999, line 65: Does this just get ID or what will this string look like

src/v2i-hub/CARMAStreetsPlugin/src/J2735MapToJsonConverter.cpp
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    void J2735MapToJsonConverter::convertLaneAttributeToJSON(const GenericLane *lane, Json::Value &LaneAttributesJson) const
+    {
+        std::stringstream ss;
+        ss.str("");
+        auto bit_unused = lane->laneAttributes.directionalUse.bits_unused;
+        ss << lane->laneAttributes.directionalUse.buf;
+        auto binary = lane->laneAttributes.directionalUse.buf[0] >> bit_unused;
+        std::string binary_str = std::to_string(static_cast<unsigned>(binary / 2)) + std::to_string(static_cast<unsigned>(binary % 2));
```
paulbourelly999, line 105: What conversion is happening here?

src/v2i-hub/CARMAStreetsPlugin/src/J2735MapToJsonConverter.cpp
dan-du-car: I can send you XML offline

src/v2i-hub/CARMAStreetsPlugin/src/J2735MapToJsonConverter.cpp
dan-du-car: LayerType is a pointer

src/v2i-hub/CARMAStreetsPlugin/src/J2735MapToJsonConverter.cpp
dan-du-car: We check the array nullable before accessing it

src/v2i-hub/CARMAStreetsPlugin/src/J2735MapToJsonConverter.cpp
dan-du-car: directionalUse is one byte. We are sending binary 01 or 10 in a string format. I can send you the XML.

src/v2i-hub/CARMAStreetsPlugin/src/J2735MapToJsonConverter.cpp
```diff
@@ -0,0 +1,267 @@
...
+            {
+                Json::Value connJson;
+                ss.str("");
+                auto connect = lane->connectsTo->list.array[i];
+                ss << connect->connectingLane.lane;
+                connJson["connecting_lane"]["lane"] = ss.str();
+
+                ss.str("");
+                connect = lane->connectsTo->list.array[i];
+                ss << *connect->signalGroup;
```
paulbourelly999, line 261: I am not clear why you sometimes need to deference the Map object or Map Object elements and other times not. Are there specific types this is necessary for?

paulbourelly999:
@dan-du-car A nice to have would be if you could add the json MAP message sample of definition to Confluence. We have definitions of the BSM, MobilityOperation, Mobility Path on this page. You could add it there. Also please feel free to update any messages you think are out of date. https://usdot-carma.atlassian.net/wiki/spaces/CRMTSMO/pages/2182873096/CARMA+Streets+Message+Data+Collection

src/v2i-hub/CARMAStreetsPlugin/src/J2735MapToJsonConverter.cpp
dan-du-car: I only dereference it when it is a pointer type. Not when it is not a pointer. This is from the J2735 MAP message definition.

src/v2i-hub/CARMAStreetsPlugin/src/J2735MapToJsonConverter.cpp
dan-du-car: When there is no laneSet at all from this MAP object, this is the checking. There is no empty function here. Only count to return the number of element. But it will throw coredump if the laneSet does not exist in the intersection, which is not the case for most valid MAP message.

src/v2i-hub/CARMAStreetsPlugin/src/J2735MapToJsonConverter.cpp
dan-du-car: Updated the confluence page with the MAP JSON, it should show the ingress_approach value with string. It is a number but in string format in JSON.

src/v2i-hub/CARMAStreetsPlugin/src/J2735MapToJsonConverter.cpp
dan-du-car: Yeah, it is a for loop to check multiple intersection. But there is only one intersection geometry in this array. So I guess this is a note to remind of the number of intersection geometry during the testing with the west intersection MAP UPER file.

paulbourelly999:
@dan-du-car I see a lot of code smells related to not making handle message methods const. Is there a reason we don't do this.

paulbourelly999 (approved):
Looks good to me.

PR #413, opened 2022-08-09
Fix optional fields

dan-du-car:
# PR Details
## Description
spat name and movement name should be optional fields

## Related Issue
NA

## Motivation and Context
UC3

## How Has This Been Tested?
unit test

## Types of changes

## Checklist:


paulbourelly999:
@dan-du-car  could we also add a try_catch similar getEncodedtsm3 method to the SubscribeSpatKafkaTopic method where we encode the SPAT to catch J2735 exceptions and log them rather than letting the plugin crash?

src/v2i-hub/CARMAStreetsPlugin/src/CARMAStreetsPlugin.cpp
```diff
@@ -588,7 +588,16 @@ void CARMAStreetsPlugin::SubscribeSpatKafkaTopic(){
 					auto spat_ptr = std::make_shared<SPAT>();
 					spat_convertor.convertJson2Spat(payload_root, spat_ptr.get());
 					tmx::messages::SpatEncodedMessage spatEncodedMsg;
-					spat_convertor.encodeSpat(spat_ptr, spatEncodedMsg);
+					try
+					{
+						spat_convertor.encodeSpat(spat_ptr, spatEncodedMsg);
+					}
+					catch(...)
```
paulbourelly999, line 595: Last thing. Is it possible to catch the specific exception since this will throw a code smell.

paulbourelly999: @dan-du-car  Why did we mark this as resolved. In some of the other methods you catch a TmxException &ex. Can we do this here as well.

Issue #415, opened 2022-09-02
CARMA-Streets Plugin Incorrectly translates BSM IDs with leading zeros


paulbourelly999:
### Types of Issue

- [x ] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

CARMA-Streets Plugin attempts to translate each byte in byte array to a hex but does not account for leading zeros for each byte and instead left pads the resulting hex number with zeros to make it 4 bytes long.

### Expected behavior
BSM id is translated to hex correctly by CARMA-Streets Plugin
### Actual behavior

Incorrect for bytes with leading zeros.

PR #419, opened 2022-09-16
Fix BSM ID

paulbourelly999:
# PR Details
## Description

Fix BSM ID translation issues
## Related Issue
#415 

## Types of changes

- [ x] Defect fix (non-breaking change that fixes an issue)

## Checklist:

### DIFF
--- a/src/v2i-hub/CARMAStreetsPlugin/src/CARMAStreetsPlugin.cpp
+++ b/src/v2i-hub/CARMAStreetsPlugin/src/CARMAStreetsPlugin.cpp
@@ -329,7 +329,7 @@ void CARMAStreetsPlugin::HandleBasicSafetyMessage(BsmMessage &msg, routeable_mes
 			 id_num = (id_num << 8) | bsm->coreData.id.buf[i];
 		}
 		std::stringstream id_fill_ss;
-		id_fill_ss << std::hex << id_num;
+		id_fill_ss << std::setfill('0') << std::setw(8) <<std::hex << id_num;
 		coreData["id"]  = id_fill_ss.str();
 		
 		Json::Value accuracy;


src/v2i-hub/CARMAStreetsPlugin/src/CARMAStreetsPlugin.cpp
```diff
@@ -329,7 +329,7 @@ void CARMAStreetsPlugin::HandleBasicSafetyMessage(BsmMessage &msg, routeable_mes
 			 id_num = (id_num << 8) | bsm->coreData.id.buf[i];
 		}
 		std::stringstream id_fill_ss;
-		id_fill_ss << std::hex << id_num;
+		id_fill_ss << std::setfill('0') << std::setw(8) <<std::hex << id_num;
```
dan-du-car, line 332: Why do we need this change?

paulbourelly999: So two fixes have been applied for BSM IDs since you have been out. This is the second fix. Here is the first fix PR that has already been closed (https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/416). The issue for the first PR is that your old code  did not seem to correctly translate bytes in the BSM id. Specifically your code seemed to translate bytes that had leading zeros (i.e. 0a) without the leading zero (i.e. a ) and append all missing zeros as the end (i.e 12340a56 => 01234a56). I fixed this by replacing your logic with the BSM ID translation logic from platform (https://github.com/usdot-fhwa-stol/carma-utils/blob/develop/bsm_helper/src/bsm_helper.cpp). This worked well for most BSM id's but seemed to fail specifically for IDs that had one left padded zero (i.e 05000000 => 5000000). Not sure why this first zero has being removed but with the additional changes I added it guarantees a 4 byte hex and left pads the result with zero as well as fixing the internal bytes to all have leading zeros if necessary. Feel free to test this solution as much as you want or suggest alternatives. For regression testing we had a lot of issues with incorrectly translated BSM ids which I am so surprised we did not run into for UC 1 testing a while ago but they all seem to be issues. Hopefully this was helpful information

PR #429, opened 2022-11-03
Fix compilation errors on Ubuntu 22.04

oschwartz10612:
When compiling on Ubuntu 22.04 the following errors prevented successful compilation: 

`HistoryManager.cpp:38 error: taking address of rvalue`
`TmxException.cpp:69 error: ordered comparison of pointer with integer zero ('void*' and 'int')`

This pull request fixes these two issues.

### DIFF
--- a/src/tmx/TmxApi/tmx/TmxException.hpp
+++ b/src/tmx/TmxApi/tmx/TmxException.hpp
@@ -66,7 +66,7 @@ public:
 		int nptrs = backtrace(btBuffer, BT_BUFFER_SIZE);
 
 		/* overwrite sigaction with caller's address */
-		if (nptrs > 1 && callerAddr > 0)
+		if (nptrs > 1 && callerAddr != 0)
 			btBuffer[1] = callerAddr;
 
 		char **symbols = backtrace_symbols(btBuffer, nptrs);
--- a/src/tmx/TmxCore/src/HistoryManager.cpp
+++ b/src/tmx/TmxCore/src/HistoryManager.cpp
@@ -34,8 +34,10 @@ HistoryManager::HistoryManager(MessageRouter *messageRouter) : Plugin(messageRou
 	info.pluginInfo.description = "Core element that is responsible for purging old log and history data";
 	info.pluginInfo.version = IVPCORE_VERSION;
 
+	ostringstream m = ostringstream() << HISTORYPURGER_CONFIGDFLT_PURGEINTERVAL;
+
 	info.configDefaultEntries.push_back(PluginConfigurationParameterEntry(HISTORYPURGER_CONFIGKEY_PURGEINTERVAL,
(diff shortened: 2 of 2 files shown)


oschwartz10612:
Checks seem to have failed on a docker CLI issue: https://app.circleci.com/pipelines/github/usdot-fhwa-OPS/V2X-Hub/1913/workflows/d706e96b-f9eb-4927-a224-04d0dd50eb56/jobs/3349?invite=true#step-102-7

rob-baily:
@codygarver The CI logs seem to indicate the problem is because the branch and tag name being used is "pull/429".  It seems this is not a valid docker tag syntax.  I am guessing this name happened because it is a forked repo but I don't know for sure.  Any ideas on whether a setting can be changed to make it "pull-429" or if we have to change the CI setup?

codygarver:
@rob-baily I've looked more closely and, as you suspect, it is because this is a forked PR. There are actually a few things going wrong:
1. the docker tag is trying to use the branch variable that has a "/" character in it, which is easy enough to solve
2. even if the tag was acceptable the CI will try to docker push without credentials
3. sonar scan relies on pulling the image pushed in step 2

So overall I'd say the CI needs a big re-think. I'm not sure what to do with this conclusion though, if it goes on the DevOps board or what. Needs V2X management opinion I guess.

rob-baily:
@oschwartz10612 How did you get around #432 first?  I did a brief test using the packaged version of libwebsockets for Ubuntu 22 and there were linking errors.

FYI we are working to get forked pull requests working in the CI through #450.

oschwartz10612:
Its been a little it sense i did this but I believe I just installed the upstream libwebsockets and used that and everything compiled fine. The issue with that if I remember correctly was that plugin uploads and a few other things were not working. I am assuming whatever patches were made in the fhwa version are necessary.

I think step 1 for 22.04 would be to get libwebsockets back in sync. I started trying to do that but ran out of time.

rob-baily:
@oschwartz10612  A couple things on this:

- It seems that even the libwebsockets version that is in the Ubuntu repo for Jammy is still not new enough as you surmised.
- The code change that you have proposed does not work on the current Ubuntu 18 platform.

We have pushed an ubuntu_22 branch with some changes that should work if you want to try it out.  Note:
- In the ext directory there is a build.sh script which should be used to build some of the other libraries used.  It gets them from tagged versions of artifacts in other libraries when the code is not included in this one.
- We took your code and modified it to work on both.  The statement that was in there before your change was a bit convoluted so it is more straightforward now.

Please see what you think and if it looks good we can merge to develop.

rob-baily:
@oschwartz10612 We went ahead and merged this in and then made the change so it can compile for Ubuntu 20 and 22.  Please go ahead and check out the latest in develop branch and see if that works for you,

oschwartz10612:
Great! I will play with it as soon as I can!

Issue #444, opened 2022-12-09
Update docker compose configuration files to allow MySQL data persistence across restarts


rob-baily:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [x] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary
Use the current docker compose example files whenever the mysql container is restarted all data is lost.  Volumes can be used to keep the database data separate from the container that it runs in.

### V2XHUB version where this issue was discovered
7.3.1

### Expected behavior
Database settings should not be lost between Docker runs.

### Actual behavior
See description.

### Steps to reproduce the actual behavior
1. Follow the current setup instructions to set everything up and bring up the system using the command "docker-compose up -d"
2. Add the use to the database.
3. Log into the system to confirm it is working.
4. Bring the system down using the command "docker-compose down"
5. Bring the system back up using the command "docker-compose up -d"
6. Attempt to log into the system using the user id/pwd combination that was configured before and it does not work.

rob-baily:
PR has been merged.

PR #445, opened 2022-12-09
Add a volume to docker compose file so that MySQL data is not dropped…

rob-baily:
… every time docker container is restarted.

# PR Details
## Description
Change docker compose example files to use volumes for MySQL data.

## Related Issue
#444 

## Motivation and Context
Data in the database should not be lost between runs.

## How Has This Been Tested?
Tested locally by myself and by Will on his setup.  Follow the steps in the linked issue and login is now possible after restart using the initially entered credentals.

## Types of changes

## Checklist:


rob-baily:
@paulbourelly999 The current documentation only mentions using "docker-compose down".  I would generally think the expectation is that database data survives restarts so we probably don't need to add anything specific on this.  We could add something to indicate using the -v option to remove the volume but I don't think it is necessary.

paulbourelly999 (approved):
Looks good.

Issue #450, opened 2022-12-13
Support CI builds for forked pull requests


codygarver:
### Types of Issue

- [ - ] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### V2XHUB version where this issue was discovered
develop

### Actual behavior

Opening a PR from a forked repo doesn't pass CI because CI tries to do a docker push with the base repo's secret

We need to...
* centralize the dependency installation to a script where possible
* use dependency install script in relevant Dockerfile and CircleCI where needed
* install Sonar only in CI
* get rid of any Dockerfile(s) only used for CI purposes
* only push desirable CI builds to dockerhub

rob-baily:
@codygarver I have pushed some changes in the branch https://github.com/usdot-fhwa-OPS/V2X-Hub/tree/ci_refactor_for_forks which I think covers the points above. One question though is that I see that we use the env SONAR_SCANNER_TOKEN during the Sonar scan.   Is that not a protected variable that would be hidden from forks?

To the points above:

- There is scripts/install_dependencies.sh which is used in the Dockerfile and CI build for the core OS dependencies.
- Sonar now no longer uses a built Docker image to run things.  I copied a number of lines from Carma Streets.  Seems like there might be some opportunity to reuse stuff, maybe in the future.
- I removed Dockerfile-depl and left just Dockerfile.
- I did not change anything for what gets pushed so maybe that needs to be updated.

One other thing is that I was not sure how to configure the Sonar job so I have it as branches but ignoring the "develop" branch.  This is probably not correct.

I did confirm that the amd image works as expected in a deployment.

Feel free to checkout the branch yourself and make any changes as necessary and I am happy to discuss anything in it.  I did not try it on a forked PR to see either so that will need to be done.

rob-baily:
I looked online and based on this thread https://community.sonarsource.com/t/code-analysis-on-pull-request-from-forked-repository-with-github-actions/43986 Sonar does not support scans on forked PRs.  It is on their roadmap.

rob-baily:
Some notes on this:

[...]

PR #454, opened 2022-12-15
Ci refactor for forks

rob-baily:
# PR Details
## Description
Modify CI configuration to handle forked pull requests which do not have access to any secret values.

## Related Issue
Addressing #450 

## Motivation and Context

Anyone submitting a PR through a fork cannot have their CI checks passed without this.

## Types of changes

## Checklist:


container/setup.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+set -ex
+
+cp /home/V2X-Hub/src/tmx/TmxCore/tmxcore.service /lib/systemd/system/
+cp /home/V2X-Hub/src/tmx/TmxCore/tmxcore.service /usr/sbin/
+chmod +x /home/V2X-Hub/container/service.sh /home/V2X-Hub/container/wait-for-it.sh
+mkdir -p /var/www/plugins/MAP /var/www/plugins/.ssl
+chown plugin:www-data /var/www/plugins/.ssl
+openssl req -x509 -newkey rsa:4096 -sha256 -nodes -keyout /var/www/plugins/.ssl/tmxcmd.key -out tmxcmd.crt -subj "/CN= <127.0.0.1> " -days 3650
+chown plugin:www-data /var/www/plugins/.ssl/*
```
rob-baily, line 10: @codygarver Looks like line 10 is a duplicate of line 8.  Probably one is support to be a chgrp as that is what it was before.

rob-baily: Also I don't think the : syntax is correct here.  Are you able to test these?

codygarver: @rob-baily that's the correct chown syntax, and it looks right in the built image. However, openssl was missing a flag (which is fixed now) and I went ahead and combined all of the chown while I was there.

codygarver (approved):

Issue #432, opened 2022-11-10
libwebsockets does not build on Ubuntu 22.04


oschwartz10612:
libwebsockets does not build on Ubuntu 22.04. See the below errors:

```
#0 12.50 /home/V2X-Hub/ext/libwebsockets/lib/tls/openssl/openssl-server.c: In function 'lws_tls_server_certs_load':
#0 12.50 /home/V2X-Hub/ext/libwebsockets/lib/tls/openssl/openssl-server.c:396:9: error: 'EC_KEY_new_by_curve_name' is deprecated: Since OpenSSL 3.0 [-Werror=deprecated-declarations]
#0 12.50   396 |         ecdh = EC_KEY_new_by_curve_name(ecdh_nid);
#0 12.50       |         ^~~~
#0 12.50 In file included from /usr/include/openssl/x509.h:33,
#0 12.50                  from /usr/include/openssl/ssl.h:31,
#0 12.50                  from /home/V2X-Hub/ext/libwebsockets/include/libwebsockets.h:229,
#0 12.50                  from /home/V2X-Hub/ext/libwebsockets/lib/core/private.h:126,
#0 12.50                  from /home/V2X-Hub/ext/libwebsockets/lib/tls/openssl/openssl-server.c:22:
#0 12.50 /usr/include/openssl/ec.h:998:31: note: declared here
#0 12.50   998 | OSSL_DEPRECATEDIN_3_0 EC_KEY *EC_KEY_new_by_curve_name(int nid);
#0 12.50       |                               ^~~~~~~~~~~~~~~~~~~~~~~~
#0 12.50 /home/V2X-Hub/ext/libwebsockets/lib/tls/openssl/openssl-server.c:402:9: error: 'EC_KEY_free' is deprecated: Since OpenSSL 3.0 [-Werror=deprecated-declarations]
#0 12.50   402 |         EC_KEY_free(ecdh);
#0 12.50       |         ^~~~~~~~~~~
#0 12.50 In file included from /usr/include/openssl/x509.h:33,
#0 12.50                  from /usr/include/openssl/ssl.h:31,
#0 12.50                  from /home/V2X-Hub/ext/libwebsockets/include/libwebsockets.h:229,
#0 12.50                  from /home/V2X-Hub/ext/libwebsockets/lib/core/private.h:126,
#0 12.50                  from /home/V2X-Hub/ext/libwebsockets/lib/tls/openssl/openssl-server.c:22:
[...]

PR #458, opened 2022-12-21
Ubuntu 22

rob-baily:
# PR Details
## Description

This PR allows V2XHub to be built and run on Ubuntu 22.  Some of the things to note in this PR:

- The 2 source code changes are very close to what was proposed in https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/429.  It turns out one of them did not compile on Ubuntu 20 so it was modified.
- The Dockerfile was updated to allow the Ubuntu version to be passed in via build argument with the default being Ubuntu 20 that was used before.
- The custom build for libwebsockets was removed as it is no longer needed.  The repository versions for Ubuntu 20 and 22 work fine.  We should remove https://github.com/usdot-fhwa-OPS/libwebsockets in the not too distant future.
- A separate Ubuntu 22 CI job was added,  This means a CI build will create a Docker image for Ubuntu 20 and 22.  Currently the 22 version is not published.
- Documentation relating to Ubuntu 16 and 18 was removed.  It is assumed Ubuntu 20 should be our go forward minimum version since 18LTS is end of lifed in early 2023.

## Related Issue

#432 

## Types of changes

## Checklist:


src/v2i-hub/package_plugins.sh
```diff
@@ -23,6 +23,8 @@ PLUGINS=(
 
 # package up all plugins
 for PLUGIN in "${PLUGINS[@]}"; do
+    # remove the link if it exists already
+    rm -f "$PLUGIN"/bin
```
codygarver, line 27: This line can be avoided by adding `--force` to `ln` command on the next line

ext/build.sh
```diff
@@ -3,27 +3,23 @@
...
-cmake -DLWS_WITH_SHARED=OFF .
-make
-make install
-popd
-
 # An OPENAPI based Qt webservice is needed by the plugins for http requests processing. A custom generated code using OPENAPI framework is located in GitHub.
 pushd /tmp
-git clone https://github.com/usdot-fhwa-OPS/qhttpengine.git
-cd qhttpengine/
+#git clone https://github.com/usdot-fhwa-OPS/qhttpengine.git
```
codygarver, line 8: Remove commented out code

docs/Installation_and_Setup.md
```diff
@@ -95,6 +91,6 @@ You can paste the map input file to this location:
 ```
 
 ## Security and Passwords
-V2X Hub is middleware that runs on Linux Ubuntu 16.04 LTS or Ubuntu 18.04 LTS with future versions on Ubuntu 18.04 LTS. It is recommended that appropriate security and firewall settings be used on the computer running Linux, including conforming to your agency's security best practices and IT protocols.
+V2X Hub is middleware that runs on Linux Ubuntu within a Docker image.  It is recommended that appropriate security and firewall settings be used on the computer running Linux, including conforming to your agency's security best practices and IT protocols.
```
codygarver, line 94: While you're changing this line please also change `Linux Ubuntu` to `Ubuntu Linux`

.circleci/config.yml
```diff
@@ -129,6 +129,16 @@ jobs:
             docker push usdotfhwaops/v2xhubamd:${DOCKER_TAG}
             docker push usdotfhwaops/php:${DOCKER_TAG}
             docker push usdotfhwaops/port-drayage-webservice:${DOCKER_TAG}
+  docker_build_ubuntu22:
```
codygarver, line 132: Is there any way we could use matrix for ubuntu versions https://circleci.com/docs/using-matrix-jobs/

codygarver (changes requested):

src/v2i-hub/package_plugins.sh
rob-baily: Changed to use --force on ln

docs/Installation_and_Setup.md
rob-baily: Updated

ext/build.sh
rob-baily: Removed

.circleci/config.yml
rob-baily: Because the jobs are not quite equivalent I think this is a bit more complex than we want to do right now.  Also I think when we get to doing multi-architecture images it will be a different setup anyhow,

codygarver (approved):

oschwartz10612:
This is great @rob-baily! I seem to recall that when I was playing with this I did something similar in using the upstream libwebsockets and it mostly worked but I had issues uploading plugins from the web page I think. Were you able to confirm that this works? 

I can do some testing myself when I have some more time.

rob-baily:
@oschwartz10612 I had not tested the plugin upload before but just did it with the ubuntu 22 version and removing and reinstalling the CARMACloud plugin and it worked as expected.  Let us know if you believe you still see issues with the latest.

PR #460, opened 2022-12-28
V2xhub  forwards the BSM messages from ERV to cloud and broadcast BSMs returned from cloud

dan-du-car:
# PR Details
## Description
As V2xHub, I will forward the BSM messages I received from  ERV to carma-cloud. Once the RSU's identified by the carma-cloud, v2xhub equipped with the identified RSU will receive the BSM messages and forward it to immediate forward plugin to broadcast BSM messages. Refer to: https://usdot-carma.atlassian.net/wiki/spaces/CRMCLD/pages/2379808813/Emergency+response+use+case+detailed+design

## Related Issue
NA

## Motivation and Context
Emergency Response Use Case

## How Has This Been Tested?
Unit test 

## Types of changes

## Checklist:

### DIFF
--- a/src/v2i-hub/ERVCloudForwardingPlugin/CMakeLists.txt
+++ b/src/v2i-hub/ERVCloudForwardingPlugin/CMakeLists.txt
@@ -1,0 +1,37 @@
+PROJECT(ERVCloudForwardingPlugin VERSION 7.3.1 LANGUAGES CXX)
+
+SET(TMX_PLUGIN_NAME "ERVCloudForwarding")
+add_compile_options(-fPIC)
+
+find_package(Qt5Core REQUIRED)
+find_package(Qt5Network REQUIRED)
+find_package(CURL REQUIRED)
+
+find_library(libasn1c .)
+
+include_directories(
+    ${Qt5Core_INCLUDE_DIRS}
+    ${Qt5Network_INCLUDE_DIRS}
+)
+
+BuildTmxPlugin()
+
+TARGET_INCLUDE_DIRECTORIES(${PROJECT_NAME} PUBLIC ${NETSNMP_INCLUDE_DIRS} ${Qt5Core_INCLUDE_DIRS} ${Qt5Network_INCLUDE_DIRS} ${CURL_INCLUDE_DIRS})
+TARGET_LINK_LIBRARIES(${PROJECT_NAME} PUBLIC tmxutils ${NETSNMP_LIBRARIES} ${QHttpEngine_LIBRARY} ${Boost_SYSTEM_LIBRARY} curl Qt5Core Qt5Network qhttpengine v2xhubWebAPI)
+link_directories(${CMAKE_PREFIX_PATH}/lib)
+
+# ############
+# # Testing ##
+# ############
+enable_testing()
+include_directories(${PROJECT_SOURCE_DIR}/src)
+add_library(${PROJECT_NAME}_lib src/ERVCloudForwardingWorker.cpp)
+target_link_libraries(${PROJECT_NAME}_lib PUBLIC ${TMXAPI_LIBRARIES}
+    ${ASN_J2735_LIBRARIES}
+    tmxutils)
+set(BINARY ${PROJECT_NAME}_test)
+file(GLOB_RECURSE TEST_SOURCES LIST_DIRECTORIES false test/*.h test/*.cpp)
+set(SOURCES ${TEST_SOURCES} WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/test)
+add_executable(${BINARY} ${TEST_SOURCES})
+add_test(NAME ${BINARY} COMMAND ${BINARY})
+target_link_libraries(${BINARY} PUBLIC ${PROJECT_NAME}_lib gtest)
--- a/src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.cpp
+++ b/src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.cpp
@@ -1,0 +1,167 @@
+#include "ERVCloudForwardingPlugin.h"
+#include "ERVCloudForwardingWorker.h"
+
+namespace ERVCloudForwardingPlugin
+{
+    ERVCloudForwardingPlugin::ERVCloudForwardingPlugin(string name) : PluginClient(name)
+    {
+        UpdateConfigSettings();
+        std::lock_guard<mutex> lock(_cfgLock);
+        AddMessageFilter<BsmMessage>(this, &ERVCloudForwardingPlugin::handleBSM);
+        // Subscribe to all messages specified by the filters above.
+        SubscribeToMessages();
+
+        std::thread webBSM_t(&ERVCloudForwardingPlugin::StartBSMWebService, this);
+        webBSM_t.detach();
+    }
+
+    void ERVCloudForwardingPlugin::handleBSM(BsmMessage &msg, routeable_message &routableMsg)
+    {
+        // Check if the BSM is broadcast by an ERV (Emergency Response Vehicle)
+        if (ERVCloudForwardingWorker::IsBSMFromERV(msg))
+        {
+            // Construct the ERV BSM and forward it to the cloud.
+            auto xml_str = ERVCloudForwardingWorker::constructERVBSMRequest(msg);
+            PLOG(logINFO) << "Forward ERV BSM to cloud: " << xml_str << endl;
+            CloudSendAsync(xml_str, _CLOUDURL, _CLOUDBSMREQ, _METHOD);
+        }
+        else
+        {
+            // If BSM is not from ERV, print debug log
+            PLOG(logDEBUG) << "Incoming BSM is not from Emergency Response Vehicle (ERV)." << endl;
+        }
+    }
+
+    void ERVCloudForwardingPlugin::UpdateConfigSettings()
+    {
+        std::lock_guard<mutex> lock(_cfgLock);
+        GetConfigValue<string>("WebServiceIP", _webIp);
+        GetConfigValue<uint16_t>("WebServicePort", _webPort);
+    }
+
+    void ERVCloudForwardingPlugin::BroadcastBSM(const string &bsmHex)
+    {
+        std::unique_ptr<BsmEncodedMessage> msg;
+        J2735MessageFactory factory;
+        msg.reset();
+        msg.reset(dynamic_cast<BsmEncodedMessage *>(factory.NewMessage(api::MSGSUBTYPE_BASICSAFETYMESSAGE_STRING)));
+        msg->set_payload(bsmHex);
+        msg->set_flags(IvpMsgFlags_RouteDSRC);
+        msg->addDsrcMetadata(172, 0x8002);
+        msg->set_encoding(_HEXENC);
+        msg->refresh_timestamp();
+        routeable_message *rMsg = dynamic_cast<routeable_message *>(msg.get());
+        BroadcastMessage(*rMsg);
+        PLOG(logDEBUG) << "Broadcast ERV BSM:" << msg->get_payload() << endl;
+    }
+
+    void ERVCloudForwardingPlugin::CARMACloudResponseHandler(QHttpEngine::Socket *socket)
+    {
+        QString st;
+        while (socket->bytesAvailable() > 0)
+        {
+            auto readBytes = socket->readAll();
+            st.append(readBytes);
+        }
+        QByteArray array = st.toLocal8Bit();
+        char *_cloudUpdate = array.data();
+        string bsmHex = _cloudUpdate;
+        PLOG(logINFO) << "Received ERV BSM from cloud:" << bsmHex << endl;
+        BroadcastBSM(bsmHex);
+    }
+
+    int ERVCloudForwardingPlugin::StartBSMWebService()
+    {
+        char *placeholderX[1] = {0};
+        int placeholderC = 1;
+        QCoreApplication a(placeholderC, placeholderX);
+        QHostAddress address = QHostAddress(QString::fromStdString(_webIp));
+        quint16 port = static_cast<quint16>(_webPort);
+        QSharedPointer<OpenAPI::OAIApiRequestHandler> handler(new OpenAPI::OAIApiRequestHandler());
+        handler = QSharedPointer<OpenAPI::OAIApiRequestHandler>(new OpenAPI::OAIApiRequestHandler());
+        auto router = QSharedPointer<OpenAPI::OAIApiRouter>::create();
+        router->setUpRoutes();
+        QObject::connect(handler.data(), &OpenAPI::OAIApiRequestHandler::requestReceived, [&](QHttpEngine::Socket *socket)
+                         { CARMACloudResponseHandler(socket); });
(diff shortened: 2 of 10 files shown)


rob-baily:
@dan-du-car I think that the sonar-scanner.properties file should be updated to include the new plugin being added.  I can guess what should be added but I am not 100% sure.   If Cody and Sai have not discussed any of this with you we should probably get them to set up a meeting to explain how devs need to keep their projects updated for Sonar.  I can try to help add some if you like and see what happens.

dan-du-car:
> @dan-du-car I think that the sonar-scanner.properties file should be updated to include the new plugin being added. I can guess what should be added but I am not 100% sure. If Cody and Sai have not discussed any of this with you we should probably get them to set up a meeting to explain how devs need to keep their projects updated for Sonar. I can try to help add some if you like and see what happens.

@rob-baily Added the plugin to sonar scanner properties waiting for the result.

scripts/install_dependencies.sh
```diff
@@ -27,9 +27,10 @@ DEPENDENCIES="build-essential \
     qtbase5-dev \
     uuid-dev \
     wget \
-    zip \
     zlib1g \
-    curl"
+    zip \
+    wget \
```
rob-baily, line 32: wget is already included above

scripts/install_dependencies.sh
```diff
@@ -27,9 +27,10 @@ DEPENDENCIES="build-essential \
...
+    libperl-dev"
```
rob-baily, line 33: What is the purpose of the perl library?

ext/build.sh
```diff
@@ -44,3 +44,22 @@ cmake .
...
+
+# SNMP library
+pushd /tmp
+wget http://sourceforge.net/projects/net-snmp/files/net-snmp/5.9.1/net-snmp-5.9.1.tar.gz
+tar -xvzf net-snmp-5.9.1.tar.gz
+cd net-snmp-5.9.1
+./configure --with-default-snmp-version="3" --with-sys-contact="@@no.where" --with-sys-location="Unknown" --with-logfile="/var/log/snmpd.log" --with-persistent-directory="/var/net-snmp"
+make
+make install
+popd
```
rob-baily, line 65: Lets make sure that if we need to use this version we remove the package install in the dependencies script.  We should not have 2 versions of the SNMP library installed as that will create confusion.

scripts/install_dependencies.sh
dan-du-car: addressed

scripts/install_dependencies.sh
dan-du-car: addressed

paulbourelly999:
@dan-du-car there is a vulnerability reported by sonar scanner. Have we investigated this. Additionally I know this is probably not what anyone wants to here but should we start enforcing the code coverage requirements for V2X-Hub since sonar scanner is apparently fixed?

src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.cpp
```diff
@@ -0,0 +1,214 @@
...
+            for (auto itr = gps_map.begin(); itr != gps_map.end(); itr++)
+            {
+                latitude = itr->first;
+                longitude = itr->second;
+            }
+
+            if (latitude == 0 || longitude == 0)
+            {
+                PLOG(logERROR) << "Invalid latitude and longitude. Cannot register RSU location." << endl;
+                return;
```
paulbourelly999, line 56: Do we want to throw exceptions here instead of returning null?

src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.cpp
```diff
@@ -0,0 +1,214 @@
...
+            }
+            auto uuid = boost::uuids::random_generator()();
+            string rsu_identifier = _rsuName + "_" + boost::lexical_cast<std::string>(uuid);
+            auto xml_str = ERVCloudForwardingWorker::constructRSULocationRequest(rsu_identifier, _webPort, latitude, longitude);
+            PLOG(logINFO) << "Sending registering RSU location reqest to cloud: " << xml_str << endl;
+            auto status = CloudSend(xml_str, _CLOUDURL, _CLOUDRSUREQ, _POSTMETHOD);
+            if (status == 1)
+            {
+                PLOG(logERROR) << "Cannot register RSU location. Reason: Failed to send RSU location to cloud." << endl;
+                return;
```
paulbourelly999, line 66: Same as above

src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.cpp
```diff
@@ -0,0 +1,214 @@
...
+            }
+        }
+        catch (SNMPClientException &ex)
+        {
+            PLOG(logERROR) << "Cannot register RSU location. Reason: " << ex.what() << endl;
+            return;
```
paulbourelly999, line 72: Same as above

src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.cpp
```diff
@@ -0,0 +1,214 @@
...
+        router->setUpRoutes();
+        QObject::connect(handler.data(), &OpenAPI::OAIApiRequestHandler::requestReceived, [&](QHttpEngine::Socket *socket)
+                         { CARMACloudResponseHandler(socket); });
+        QObject::connect(handler.data(), &OpenAPI::OAIApiRequestHandler::requestReceived, [&](QHttpEngine::Socket *socket)
+                         { router->processRequest(socket); });
+        QHttpEngine::Server server(handler.data());
+        if (!server.listen(address, port))
+        {
+            qCritical("Unable to listen on the specified port.");
+            return 1;
```
paulbourelly999, line 140: Is there a reason why we return an int for this?

src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.cpp
dan-du-car: I prefer to log the error and not end the execution by throwing the exception.

src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.cpp
dan-du-car: The same idea as the above

src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.cpp
dan-du-car: The same idea as the above

src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.cpp
dan-du-car: I think 1 represent failure.

ext/build.sh
```diff
@@ -44,3 +44,12 @@ cmake .
 make
 make install
 popd
+
+# GPS Parser
+pushd /tmp
+git clone https://github.com/ckgt/NemaTode.git
```
paulbourelly999, line 50: @rob-baily in the new updated build process should we be cloning an entire git repo or just pull a tar and build that?

src/v2i-hub/ERVCloudForwardingPlugin/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
+PROJECT(ERVCloudForwardingPlugin VERSION 7.3.1 LANGUAGES CXX)
+
+SET(TMX_PLUGIN_NAME "ERVCloudForwarding")
+add_compile_options(-fPIC)
+
+find_package(Qt5Core REQUIRED)
+find_package(Qt5Network REQUIRED)
+find_package(CURL REQUIRED)
```
paulbourelly999, line 8: Is there a reason we are using Qt core our REST API server and then CURL for our REST client. Not sure this is the correct place to address this but we would like to start consolidate the libraries we use. Will leave it up to @dan-du-car if we think tackling this issue as part of this PR is appropriate or we should try to improve this in the future.

ext/build.sh
rob-baily: Odd that I did not notice this before.  This is not ideal but it does not seem like this repo has any releases.  We could try to make sure we get a current commit based on hash that we know is good.

src/v2i-hub/ERVCloudForwardingPlugin/CMakeLists.txt
dan-du-car: I would leave it for future improvement and create a story/issue specifically to address that if needed.

rob-baily: Paul, QT is not for client applications, just server.  I think the real consolidation we should make at some point is removing the use of the curl executable and using the C++ library for that.  The CURL library is already used the CARMACloud plugin.  I think this is fine as is.

ext/build.sh
rob-baily: Thats what I found as well.  I think this is fine as is for now.

paulbourelly999: Sounds good.

src/v2i-hub/ERVCloudForwardingPlugin/CMakeLists.txt
paulbourelly999: For the record QT has both as Client and Server functionality. We have used both in CARMA-Streets and in V2X-Hub (See Port Drayage Plugin). The one benefit of using QT is that some OpenAPI codgen tools support generation of QT REST Clients and Servers based on an OpenAPI api description doc. If we do not want to worry about this here then that is fine although every additional implementation we make of a REST Client is more work for us to clean up in the future. We have also used CURL in multiple places in V2X-Hub and it seems to be less lines of code so at some point we need to make a decision and move from there

src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.cpp
paulbourelly999: You can catch the error in the calling function and handle it in some specific way. I think like this, the calling function would have no indication of whether an error occurred or not so it would not be possible to write any custom handling code for these cases.

src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.h
```diff
@@ -0,0 +1,97 @@
...
+    public:
+        explicit ERVCloudForwardingPlugin(const string &);
+        ~ERVCloudForwardingPlugin() = default;
+        int Main();
+
+    protected:
+        void UpdateConfigSettings();
+        void OnConfigChanged(const char *key, const char *value);
+        void OnStateChange(IvpPluginState state);
+        int StartBSMWebService();
```
paulbourelly999, line 70: Can we add a description to this method to indicate 1 represents failure to start web service.

src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.h
```diff
@@ -0,0 +1,97 @@
...
+        /**
+         * @brief Send BSM request to the cloud
+         * @param msg BSM message request
+         * @param url The server IP and port the request is sent to
+         * @param base The request path
+         * @param method HTTP method POST
+         * @return int http status indicator
+         */
+        int CloudSend(const string &msg, const string &url, const string &base, const string &method);
+        void CloudSendAsync(const string &msg, const string &url, const string &base, const string &method);
```
paulbourelly999, line 87: Could we add comment for this method

src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.cpp
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    void ERVCloudForwardingPlugin::RegisterRSULocation()
+    {
+        try
+        {
+            PLOG(logINFO) << "Create SNMP Client to connect to RSU. RSU IP:" << _rsuIp << ",\tRSU Port:" << _snmpPort << ",\tSecurity Name:" << _securityUser << ",\tAuthentication Passphrase: " << _authPassPhrase << endl;
+            auto snmpClient = std::make_shared<SNMPClient>(_rsuIp, _snmpPort, _securityUser, _authPassPhrase);
+            auto gps_sentense = snmpClient->SNMPGet(_GPSOID);
+            auto gps_map = ERVCloudForwardingWorker::ParseGPS(gps_sentense);
```
paulbourelly999, line 44: on exception ParseGPS returns an empty MAP. Do we need to check for that here and log something?

src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.cpp
dan-du-car: yeah, I think this is the calling function itself. The exception usually throw from other class other than inside the plugin class itself.

src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.cpp
dan-du-car: The caller will check if returned map has data.

src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.h
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/460/commits/5583c5a5a6250f44dc1acd789ae119a25e6d5462

src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.h
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/460/commits/5583c5a5a6250f44dc1acd789ae119a25e6d5462

paulbourelly999 (approved):

src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.cpp
jwillmartin: @dan-du-car Sorry, I'm late on this. Everything else looks good, but this is throwing errors from SNMPCleint() in SNMPClient.cpp, but it eventually is able to connect. Doesn't affect functionality, but do we want to look at fixing it?

dan-du-car: It is valid error if it does not connect.

Issue #331, opened 2022-02-18
Use docker multi-platform cross-compiler for AMD64 and ARM64 deployment 


paulbourelly999:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [ x ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

Hello,

Please refer to such URL: https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

I guess it would be better to maintain the releases with same Image name for different platforms, thru this 'multi-platform, cross-compiler' method.

That is, the image name could be unified to be same name, e.g. 'v2xhub' (no 'v2xhubamd', no 'v2xhubarm') and docker pull correct image type via 'platform' (or host's architecture); similar for 'php' and 'port-drayage-webservice' images.

FYR, thank you.
### Expected behavior
NA
### Actual behavior

NA
### Steps to reproduce the actual behavior

NA
### Related work

NA

rob-baily:
I tried a brief test using the branch https://github.com/usdot-fhwa-OPS/V2X-Hub/tree/multi_arch_build which is using an emulator to build the arm image.  This timed out so this solution looks unworkable. 

It looks this would need to be handled by a longer term approach of:

- Compiling the items needed and create artifacts in CI (maybe Debian packages) for them rather than compiling in place and including everything in the deployed image.
- Creating docker images using the CI created artifacts and whatever run-time libraries need to be installed.
- Splitting up the Dockerfile as mentioned in the link above and enabling cross compiling on the C and C++ classes.

jwillmartin:
Addressed with:
VH-1295
VH-1319

PR #463, opened 2022-12-30
Separate build directory

rob-baily:
# PR Details
## Description

This PR changes the CMake build process so all build generated files are put in a separate "build" directory and not intermingled with the source.  It also modifies the plugin build process to create the zip file and install it in it /usr/local/plugins to allow this directory to be scanned for the available plugins rather than keeping a hardcoded list.

Additionally the coverage report setup has been modified so if a new module with tests is added it will be automatically added into the reports.  Previously a developer would have to remember to add an entry in the Sonar config file.

## Motivation and Context

This is a better practice and keeps the source directories clean and allows for an easy "clean" operation when the build files are in one location.  It also sets the stage for easier cross compiling for use in #331.

## Types of changes

- [X ] New feature (non-breaking change that adds functionality)

## Checklist:


src/v2i-hub/CMakeLists.txt
```diff
@@ -57,6 +57,16 @@ FIND_PACKAGE (${TMX_APPNAME}-plugin
 
 find_package(Boost COMPONENTS system thread log chrono atomic regex filesystem program_options log_setup) 
 
+function(create_zip output_file input_files working_dir)
```
paulbourelly999, line 60: is this replacing the package plugin script?

paulbourelly999 (approved):
Looks good to me

rob-baily:
It turns out that packing the manifest.json was not quite right due to an order issue but that has been corrected.

PR #487, opened 2023-02-24
Arm64 fix

jwillmartin:
# PR Details
## Description
Added some changes to allow for Docker to be installed on different Linux distros for arm64.

## Motivation and Context
I wanted to be able to install Docker on both Ubuntu and Debian for arm64 PCs. 

## How Has This Been Tested?
Tested on both Ubuntu and Debian for correct script results. 

## Types of changes

## Checklist:

### DIFF
--- a/configuration/arm64/mysql/add_v2xhub_user.bash
+++ b/configuration/arm64/mysql/add_v2xhub_user.bash
@@ -3,7 +3,7 @@
 #set -e
 
 # Ensure mysql-client is installed
-REQUIRED_PKG="mysql-client"
+REQUIRED_PKG="mariadb-client-10.5"
 PKG_OK=$(dpkg-query -W --showformat='${Status}\n' $REQUIRED_PKG|grep "install ok installed")
 echo Checking for $REQUIRED_PKG: $PKG_OK
 if [ "" = "$PKG_OK" ]; then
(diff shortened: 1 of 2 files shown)


configuration/arm64/mysql/add_v2xhub_user.bash
```diff
@@ -3,7 +3,12 @@
 #set -e
 
 # Ensure mysql-client is installed
-REQUIRED_PKG="mysql-client"
+arch=$(dpkg --print-architecture)
+if [ $arch = "amd64" ]; then
```
rob-baily, line 7: Why are we checking for amd64 architecture for the arm64 configuration?  I would expect that this configuration is only used on arm architectures.  Maybe the check was supposed to be arm64?

rob-baily:
@jwillmartin Is there any background or other information on this?  I reviewed but for some stuff I don't have any context to know  what the change is for.

configuration/arm64/mysql/add_v2xhub_user.bash
jwillmartin: Ah, my bad. The idea at the moment was to merge both configurations into one folder, but I couldn't figure out how to get different docker images pulled from one docker-compose.yml file. Probably better to keep them as is.

jwillmartin:
@rob-baily We used to mostly use some single-board computer that ran a version of Ubuntu. That's why this configuration file was made to pull the Ubuntu version for Docker. But now we're using Raspberry Pi's running Debian, so we need to be able to install the correct Docker version.

rob-baily (approved):

PR #496, opened 2023-03-17
Docker build

jwillmartin:
# PR Details
## Description
These changes are meant to reduce the docker size by introducing multi-stage builds. These changes implement a build stage and a final image stage. The build stage builds everything needed to run V2X Hub. Finally, those required files and directories are copied into the final image, removing unnecessary files from the build. 

## Motivation and Context
Reducing the size of the docker image creates a more efficient tool, taking up less storage space and reducing image download time. 

## How Has This Been Tested?
Built and tested locally, then pushed/pulled to/from docker hub and tested. Ensured plugins maintained functionality and required tools were present in the image. 

## Types of changes

## Checklist:


Dockerfile
```diff
@@ -28,6 +29,30 @@ RUN ldconfig
...
+RUN ./library.sh
+RUN ldconfig
+
+COPY --from=dependencies /lib/systemd/system/tmxcore.service /lib/systemd/system/
+COPY --from=dependencies /usr/local/plugins/ /usr/local/plugins/
+COPY --from=dependencies /usr/local/lib/ /usr/local/lib/
+COPY --from=dependencies /usr/local/bin/ /usr/local/bin/
+COPY --from=dependencies /usr/sbin/tmxcore.service /usr/sbin/
+COPY --from=dependencies /var/www/plugins/ /var/www/plugins/
+COPY --from=dependencies /var/log/tmx/ /var/log/tmx/
```
paulbourelly999, line 51: You are missing some dependencies here I think. I know some tmx utils stuff is installed in `/usr/local/share/tmx/`. Additionally, all new libraries that Rob and Cody worked on get installed in `opt/carma/cmake`. Lastly I think some modules from the `ext/` directory like the pedestrian REST api  are installed in `/usr/local/share/qserverPedestrian/cmake` and `/usr/local/share/v2xhubWebAPI/cmake/`. You can try running the pedestrian plugin and see if you encounter issues to confirm this.

container/ssl.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+set -ex
+
+mkdir -p /var/www/plugins/MAP /var/www/plugins/.ssl
+openssl req -x509 -newkey rsa:4096 -sha256 -nodes -keyout /var/www/plugins/.ssl/tmxcmd.key -out /var/www/plugins/.ssl/tmxcmd.crt -subj "/CN= <127.0.0.1> " -days 3650
```
paulbourelly999, line 5: Is this script new? Does this allow for https access to the web UI or what is it's purpose.

Dockerfile
paulbourelly999: It might be easiest to just copy /opt/carma and /usr/local/share/

Dockerfile
```diff
@@ -28,6 +29,30 @@ RUN ldconfig
 
 WORKDIR /home/V2X-Hub/
 RUN /home/V2X-Hub/container/setup.sh
+
+# run final image
+FROM ubuntu:$UBUNTU_VERSION AS v2xhub
+ENV DEBIAN_FRONTEND=noninteractive
+ADD scripts/install_dependencies.sh /usr/local/bin/
```
paulbourelly999, line 36: I am not sure I understand this at all. If you are trying to setup build stages, my understanding is that you would install dependencies in the first build stage and only copy over the installed dependencies. Not sure how this makes sense if you install dependencies in both images. It looks like most of the saved memory is from just not including V2X-Hub core source files

jwillmartin: Most of the dependencies in the script are still used. I think cmake and git wouldn't be needed anymore, but the rest look like they're still used

Dockerfile
jwillmartin: Didn't run into any issues with the Ped plugin. Worked with both the websocket and rest configurations. Looking through this again to make sure I didn't miss any dependencies.

container/ssl.sh
jwillmartin: addressed

Dockerfile
jwillmartin: addressed

Dockerfile
jwillmartin: @paulbourelly999 Addressed all of these. Let me know what you think.

Dockerfile
paulbourelly999: It still looks like we are running the install dependencies twice. Am I mistaken?

paulbourelly999 (approved):
Looks good. Will merge if you have tried running the image at all.

Issue #525, opened 2023-05-04
CDASimAdapter throws exception on attempting CDASim handshake


paulbourelly999:
### Summary

On startup or disable->enable the CDASimAdapterPlugin will attempt to perform a handshake with CARMA-Simulation. To perform this handshake the plugin reads several environment variables and cast them to strings and integers. Part of this logic includes calls to the `std::stoul` function which casts a string to a long. When attempting to execute this logic the CDASimAdapter plugin will throw an exception:
```
2023-04-26 09:41:47.433316	WARNING	/home/V2X-Hub/src/tmx/TmxCore/src/logger.cpp, 117	Adding event log from 'CDASimAdapter' --> CDASimAdapter encountered unhandled exception: (Unknown source): stoul
backtrace: (Hint: Use addr2line -C -e <exe> 0x#######) to find line number)
CDASimAdapter(+0x1b665) [0x55e9f2390665]
CDASimAdapter(+0x5c60d) [0x55e9f23d160d]
CDASimAdapter(+0x5cdbb) [0x55e9f23d1dbb]
CDASimAdapter(+0x5a469) [0x55e9f23cf469]
CDASimAdapter(+0x4daf6) [0x55e9f23c2af6]
CDASimAdapter(+0x4cdc8) [0x55e9f23c1dc8]
/usr/local/lib/libtmxapi.so(ivp_onStateChange+0x82) [0x7f81357c49a4]
/usr/local/lib/libtmxapi.so(ivp_onMessageReceived+0x242) [0x7f81357c4bee]
/usr/local/lib/libtmxapi.so(ivp_receive+0x372) [0x7f81357c508a]
/lib/x86_64-linux-gnu/libc.so.6(+0x94b43) [0x7f81352cab43]
/lib/x86_64-linux-gnu/libc.so.6(clone+0x44) [0x7f813535bbb4]
diagnostic info:
/home/V2X-Hub/src/tmx/TmxUtils/src/PluginExec.cpp(188): Throw in function ExceptionToString
Dynamic exception type: tmx::utils::PluginException
std::exception::what: stoul
```

### Commit

0b57dac8eaa7fc631088a36df2fa41eb496d4c8c

### Expected Behavior

CDASimAdapterPlugin when configured correctly is able to perform handshake and maintain CDASim connection without throwing exceptions.

### Actual Behavior

CDASimAdapterPlugin fails to perform handshake since it throws an exception before sending the handshake.

[...]

PR #526, opened 2023-05-04
Fix for development image throwing exception unknown source stoul

paulbourelly999:
+ Added /usr/lib and /usr/bin to deployed image
+ Adding log statement for handshake
+ Removing cpprest library dependency which is unused

# PR Details
## Related Issue
#525 

## Motivation and Context

Fix CDASimAdapterPlugin functionality in development image
## How Has This Been Tested?

Locally integration tested using steps in the related issue.
## Types of changes

## Checklist:

- [x ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,6 +42,8 @@ RUN ldconfig
 COPY --from=dependencies /usr/local/plugins/ /usr/local/plugins/
 COPY --from=dependencies /usr/local/lib/ /usr/local/lib/
 COPY --from=dependencies /usr/local/bin/ /usr/local/bin/
+COPY --from=dependencies /usr/lib/ /usr/lib/
+COPY --from=dependencies /usr/bin/ /usr/bin/
 COPY --from=dependencies /usr/local/share/ /usr/local/share/
 COPY --from=dependencies /var/www/plugins/ /var/www/plugins/
 COPY --from=dependencies /var/log/tmx/ /var/log/tmx/
--- a/src/v2i-hub/CARMACloudPlugin/CMakeLists.txt
+++ b/src/v2i-hub/CARMACloudPlugin/CMakeLists.txt
@@ -30,7 +30,7 @@ BuildTmxPlugin ()
 
 TARGET_INCLUDE_DIRECTORIES ( ${PROJECT_NAME} PUBLIC ${XercesC_INCLUDE_DIRS} ${NETSNMP_INCLUDE_DIRS} ${Qt5Core_INCLUDE_DIRS} ${Qt5Network_INCLUDE_DIRS} ${CURL_INCLUDE_DIRS})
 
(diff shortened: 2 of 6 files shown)


Dockerfile
```diff
@@ -42,6 +42,8 @@ RUN ldconfig
 COPY --from=dependencies /usr/local/plugins/ /usr/local/plugins/
 COPY --from=dependencies /usr/local/lib/ /usr/local/lib/
 COPY --from=dependencies /usr/local/bin/ /usr/local/bin/
+COPY --from=dependencies /usr/lib/ /usr/lib/
```
dan-du-car, line 45: Do we have some lib from this folder that is relevant to this PR?

src/v2i-hub/CDASimAdapter/src/CDASimConnection.cpp
```diff
@@ -193,6 +193,8 @@ namespace CDASimAdapter{
 
     void CDASimConnection::forward_message( const std::string &msg, const std::shared_ptr<UdpClient> _client ) const {
         if ( !msg.empty() && _client) {
+            PLOG(logDEBUG) << "Sending UDP msg " << msg << " to host " << _client->GetAddress()
```
dan-du-car, line 196: Is msg is empty, do we want to log a warning or debug statement as well for later debug purpose?

Dockerfile
paulbourelly999: Some libraries installed using apt-get either install in /usr/local/ or /usr/ in my experience (not sure how this is determined but I think it's some part of the debian package setup). When we where only copying /usr/local/ our CDASimAdapter plugin would throw an 
(unknown source ) stoul exception on lines of code that used std::stoul (See related issue for this PR). After adding this the error went away. I am assuming some part of the C++ librarywas included under /usr/lib/ during the build  but then did not make it into the next run image. It still looks to me that it should have worked since we install the same library anyways using the scripts but something didn't work. I put a issue up to further fix this Docker image since right now it is a little weird (https://github.com/usdot-fhwa-OPS/V2X-Hub/issues/527)

src/v2i-hub/CDASimAdapter/src/CDASimConnection.cpp
paulbourelly999: Good Idea. Will add that.

kjrush (approved):
No additional comments beyond what Dan already mentioned!

src/v2i-hub/CARMACloudPlugin/CMakeLists.txt
jwillmartin, line 1: Will cpprest not be needed for the CARMACloud Plugin?

paulbourelly999: No CARMACloud using curl. @dan-du-car This is correct right? CARMA Clould using CURL for all its REST communication.

dan-du-car: Yeah, it used CURL

dan-du-car (approved):

PR #433, opened 2022-11-14
Immediate Forward Plugin Configurable Channel 

jwillmartin:
modified:   src/v2i-hub/DsrcImmediateForwardPlugin/manifest.json
	modified:   src/v2i-hub/DsrcImmediateForwardPlugin/src/DsrcMessageManagerPlugin.cpp

# PR Details
## Description

- Made changes to Immediate Forward plugin manifest file, including channel number option in the default parameters for message_destination_1.
- Added a note to the manifest file, indicating users to change the channel to the one needed.
- Additionally, fixed typo in code, where the message frame included an unneeded space.

## Related Issue

https://github.com/usdot-fhwa-OPS/V2X-Hub/issues/427

## Motivation and Context

This change allows users to send messages to multiple RSUs transmitting on different channels

## How Has This Been Tested?

Tested locally and decoded messages forwarded from V2X Hub to the RSU, using wireshark to verify the changes. 
Uploaded to github and verified desired changes reflected in the automatically generated docker image.
Finally, tested functionality with both a DSRC RSU on channel 172 and a C-V2X RSU on channel 183. 

## Types of changes

- [x] Defect fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [x] Breaking change (fix or feature that cause existing functionality to change)

## Checklist:

- [ ] I have added any new packages to the sonar-scanner.properties file
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
[V2XHUB Contributing Guide](https://github.com/usdot-fhwa-OPS/V2X-Hub/blob/develop/Contributing.md) 
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #545, opened 2023-06-26
Routable metadata

jwillmartin:
# PR Details
## Description
Adds PSID metadata to BSMs and SRMs that are sent to the Message Receiver plugin. If "RouteMessage" is enabled, the messages will be available for the Immediate Forward plugin to use. 

Additional changes were made to build process to speed up process.

## How Has This Been Tested?
Sent BSMs and SRMs to Message Receiver plugin. Messages were transmitted by Immediate Forward plugin when "RouteMessage" was enabled and not routed when disabled. Previously, messages were skipped or not routed, respectively. 

## Types of changes

## Checklist:

### DIFF
--- a/src/build.sh
+++ b/src/build.sh
@@ -18,6 +18,8 @@ set -e
 # script executes all tmx and v2i build and coverage steps so that they can be singularly
 # wrapped by the sonarcloud build-wrapper
 
+numCPU=$(nproc)
+
 RELEASE_BUILD=0
 if [ "$1" = "release" ]; then
     RELEASE_BUILD=1
@@ -36,7 +38,7 @@ fi
 pushd tmx
 cmake -Bbuild -DCMAKE_PREFIX_PATH=\"/usr/local/share/tmx\;\/opt/carma/cmake\;\" -DCMAKE_CXX_FLAGS="${COVERAGE_FLAGS}" -DCMAKE_C_FLAGS="${COVERAGE_FLAGS}" -DCMAKE_BUILD_TYPE="${BUILD_TYPE}" .
 pushd build
-make
+make -j${numCPU}
 make install
 popd
 popd
@@ -44,6 +46,7 @@ popd
 pushd v2i-hub
 cmake -Bbuild  -DCMAKE_PREFIX_PATH=\"/usr/local/share/tmx\;\/opt/carma/cmake\;\" -DqserverPedestrian_DIR=/usr/local/share/qserverPedestrian/cmake -Dv2xhubWebAPI_DIR=/usr/local/share/v2xhubWebAPI/cmake/ -DCMAKE_CXX_FLAGS="${COVERAGE_FLAGS}" -DCMAKE_C_FLAGS="${COVERAGE_FLAGS}" -DCMAKE_BUILD_TYPE="${BUILD_TYPE}" .
 pushd build
+make -j${numCPU}
 make install
 popd
 popd
--- a/src/tmx/TmxApi/tmx/IvpMessage.h
+++ b/src/tmx/TmxApi/tmx/IvpMessage.h
@@ -20,8 +20,8 @@ typedef unsigned int IvpMsgFlags;
 #define IvpMsgFlags_RouteDSRC 0x01
 
 typedef struct IvpDsrcMetadata {
-	int channel;
 	int psid;
+	int channel;
 } IvpDsrcMetadata;
 
 typedef struct IvpMessage {
@@ -64,7 +64,7 @@ typedef enum {
  */
 IvpMessage *ivpMsg_create(const char *type, const char *subtype, const char *encoding, IvpMsgFlags flags, cJSON *payload);
 
-IvpMessage *ivpMsg_addDsrcMetadata(IvpMessage *msg, int channel, int psid);
+IvpMessage *ivpMsg_addDsrcMetadata(IvpMessage *msg, int psid, int channel);
 
 /*!
  * Creates a new IvpMessage from a json string.
--- a/src/v2i-hub/MessageReceiverPlugin/manifest.json
+++ b/src/v2i-hub/MessageReceiverPlugin/manifest.json
@@ -29,9 +29,9 @@
 			"description":"Port for the incoming message network connection."
 		},
 		{
-			"key":"RouteDSRC",
+			"key":"RouteMessage",
 			"default":"false",
-			"description":"Set the flag to route a received J2735 message."
+			"description":"Set the flag to route/broadcast a received J2735 message to TMX Core."
 		},
 		{
 			"key":"EnableSimulatedBSM",
--- a/src/v2i-hub/MessageReceiverPlugin/manifest.json
+++ b/src/v2i-hub/MessageReceiverPlugin/manifest.json
@@ -29,7 +29,7 @@
 			"description":"Port for the incoming message network connection."
 		},
 		{
-			"key":"RouteMessage",
+			"key":"RouteJ2735",
 			"default":"false",
 			"description":"Set the flag to route/broadcast a received J2735 message to TMX Core."
 		},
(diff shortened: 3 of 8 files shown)


src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
```diff
@@ -263,6 +263,7 @@ void MessageReceiverPlugin::OnMessageReceived(routeable_message &msg)
 								}
 
 								sendMsg = encode(encodedBsm, bsm);
+								msgPSID = 0x20;
```
dan-du-car, line 266: Do different messages have different msgPSID? Does a PSID ever change for a message type? I am not sure if we want to hardcode these  values for different msgPSID?

src/build.sh
```diff
@@ -36,14 +38,15 @@ fi
...
-make
+make -j${numCPU}
 make install
 popd
 popd
 
 pushd v2i-hub
 cmake -Bbuild  -DCMAKE_PREFIX_PATH=\"/usr/local/share/tmx\;\/opt/carma/cmake\;\" -DqserverPedestrian_DIR=/usr/local/share/qserverPedestrian/cmake -Dv2xhubWebAPI_DIR=/usr/local/share/v2xhubWebAPI/cmake/ -DCMAKE_CXX_FLAGS="${COVERAGE_FLAGS}" -DCMAKE_C_FLAGS="${COVERAGE_FLAGS}" -DCMAKE_BUILD_TYPE="${BUILD_TYPE}" .
 pushd build
+make -j${numCPU}
```
dan-du-car, line 49: Why do we need this ${numCPU}? Whese is this env variable from?

src/v2i-hub/MessageReceiverPlugin/manifest.json
```diff
@@ -29,9 +29,9 @@
 			"description":"Port for the incoming message network connection."
 		},
 		{
-			"key":"RouteDSRC",
+			"key":"RouteMessage",
```
dan-du-car, line 32: Why renaming this boolean?

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
jwillmartin: These two I added were under the cases for BSMs and SRMs that are detected by the plugin. Others are kept at null. 
If we want to add more messages to be routable, we can think about where to add their PSIDs. Right now it's added within the specific plugins that generate the messages.

src/build.sh
jwillmartin: The env is for finding the number or cores available on the local machine. This allows for building in parallel. Since we're using Ubuntu as the docker image, this command to get the env variable will always work.

src/v2i-hub/MessageReceiverPlugin/manifest.json
dan-du-car: Is this boolean only for enable route for J2735Message? Some other internal v2xhub message like State, log on the UI are not controlled by this boolean? Maybe narrow down the Message to a specific type?

src/build.sh
dan-du-car: Do you have a link explain how this command "make -j${numCPU}" allow building in parallel. Sorry, I have not seen this before and am curious how this works. Also if we are running this inside a base image ubuntu, how does that get the host machine hardware specs, like num of CPUs?

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
dan-du-car: Do we have a enum or struct to consolidate all the PSID? Then use it here in this class? It might be easy to read.

src/build.sh
jwillmartin: Yep! Section 5.4 here: https://www.gnu.org/software/make/manual/make.html#Parallel

Yeah, it's still linked to the host machine, so it's able to find the number of cores available. Then this would run one job per core.

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
jwillmartin: We don't, but we could make one. What do you think about adding it here? https://github.com/usdot-fhwa-OPS/V2X-Hub/blob/develop/src/tmx/TmxApi/tmx/TmxApiMessages.h

src/v2i-hub/MessageReceiverPlugin/manifest.json
jwillmartin: Nope, just here. It's specific to some logic for broadcasting messages received by this plugin. 
I could change it to RouteJ2735

dan-du-car: make sense to me

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
dan-du-car: yep, I think that is a good location

src/build.sh
dan-du-car: If there is nothing looking like an integer after the ‘-j’ option, there is no limit on the number of job slots.

dan-du-car: image

jwillmartin: Oh, numCPU set by `numproc`

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
```diff
@@ -280,6 +281,7 @@ void MessageReceiverPlugin::OnMessageReceived(routeable_message &msg)
 										ntohl(*((uint32_t*)&(bytes.data()[24]))),
 										ntohl(*((uint32_t*)&(bytes.data()[28]))));
 								sendMsg = encode(encodedSrm, srm);
+								msgPSID = 0xE0000016;
```
paulbourelly999, line 284: There is a code smell on this line about an implicit conversion. Please check this out in the sonar scanner report.

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
jwillmartin: @dan-du-car Changes made

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
jwillmartin: @paulbourelly999 Addressed

src/tmx/TmxApi/tmx/IvpMessage.h
```diff
@@ -64,7 +64,7 @@ typedef enum {
  */
 IvpMessage *ivpMsg_create(const char *type, const char *subtype, const char *encoding, IvpMsgFlags flags, cJSON *payload);
 
-IvpMessage *ivpMsg_addDsrcMetadata(IvpMessage *msg, int channel, int psid);
+IvpMessage *ivpMsg_addDsrcMetadata(IvpMessage *msg, int psid, int channel);
```
paulbourelly999, line 67: Is there a reason we reversed these and can we ensure this does not have any unintentional impacts.

jwillmartin: @paulbourelly999 Did it to match the changes made here:
https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/433/files#diff-ea7aa329ba2d3d9e2dbe5e75039c3eadbb3886943f4596c9e0fbacb5e8af8b0c 

Did some quick tests with SPaT, MAP, PSM, and TIM to make sure they were still getting the right PSIDs set

paulbourelly999 (approved):
Looks good.

Issue #543, opened 2023-06-22
CARMA Streets Plugin Kafka Consumers


paulbourelly999:
### Summary

In the CARMA-Streets Plugin, kafka consumer threads are written to attempt to subscribe and consume from topics on configuration update events. On V2X-Hub start-up of a plugin, these update events can fire multiple times. Duplicate attempts to subscribe to the same topic can lead to rebalancing events on the kafka broker (depending on the number of partitions configured). Most of the time these rebalancing events result in small initial delays to message consumption but in CDASim testing have also been observed to hault all message consumption from Kafka by the CARMA-Streets plugin. This occurs when kafka fails to rebalance after receiving redundant subscribe calls.

### Version

4.3.0 (Current)

### Expected Behavior

CARMA Streets V2X-Hub plugin reliably consumes Kafka message from Kafka broker by correctly subscribing to topics.

### Actual Behavior

CARMA Streets V2X-Hub plugin sends duplicate subscription attempts the Kafka broker on startup due to update events fired as plugin starts up. This can cause some delay to initial message consumption and occassion no message consumptions which requires a restart to the plugin. 

### Steps to Reproduce the Actual Behavior

1. Launch CDASim new XIL release repeatedly
2. Check if SPaT messages are being received by CARMA Streets
3. If not you are likely experiencing this issue.

PR #547, opened 2023-06-30
CARMA Streets Plugin Kafka Consumers

dan-du-car:
# PR Details
## Description
- Remove the kafka consumer/producer initialization during config parameter update. 

- Replace consumer creation with v2xhub kafka_client library.
- Add producer creation in kafka_client library
## Related Issue
https://github.com/usdot-fhwa-OPS/V2X-Hub/issues/543

## Motivation and Context
NA

## How Has This Been Tested?
Local integration testing

## Types of changes

## Checklist:

### DIFF
--- a/.devcontainer/docker-compose-vscode.yml
+++ b/.devcontainer/docker-compose-vscode.yml
@@ -72,7 +72,7 @@ services:
       DOCKER_HOST_IP:  127.0.0.1
       KAFKA_ADVERTISED_HOST_NAME:  127.0.0.1
       KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
-      KAFKA_CREATE_TOPICS: "time_sync:1:1,modified_spat:1:1,v2xhub_scheduling_plan_sub:1:1"
+      KAFKA_CREATE_TOPICS: "time_sync:1:1,modified_spat:1:1,v2xhub_scheduling_plan_sub:1:1,v2xhub_ssm_sub:1:1"
       KAFKA_LOG_DIRS: "/kafka/kafka-logs"
       KAFKA_OFFSETS_RETENTION_MINUTES: 1
     volumes:
--- a/.devcontainer/docker-compose-vscode.yml
+++ b/.devcontainer/docker-compose-vscode.yml
@@ -74,7 +74,6 @@ services:
       KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
       KAFKA_CREATE_TOPICS: "time_sync:1:1,modified_spat:1:1,v2xhub_scheduling_plan_sub:1:1,v2xhub_ssm_sub:1:1"
       KAFKA_LOG_DIRS: "/kafka/kafka-logs"
-      KAFKA_OFFSETS_RETENTION_MINUTES: 1
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
       - /etc/localtime:/etc/localtime:ro
--- a/.devcontainer/docker-compose-vscode.yml
+++ b/.devcontainer/docker-compose-vscode.yml
@@ -72,8 +72,9 @@ services:
       DOCKER_HOST_IP:  127.0.0.1
       KAFKA_ADVERTISED_HOST_NAME:  127.0.0.1
       KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
-      KAFKA_CREATE_TOPICS: "time_sync:1:1"
+      KAFKA_CREATE_TOPICS: "time_sync:1:1,modified_spat:1:1,v2xhub_scheduling_plan_sub:1:1"
       KAFKA_LOG_DIRS: "/kafka/kafka-logs"
+      KAFKA_OFFSETS_RETENTION_MINUTES: 1
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
       - /etc/localtime:/etc/localtime:ro
--- a/src/tmx/TmxUtils/src/kafka/kafka_client.h
+++ b/src/tmx/TmxUtils/src/kafka/kafka_client.h
@@ -15,6 +15,7 @@ namespace tmx::utils
         std::shared_ptr<kafka_consumer_worker> create_consumer(const std::string &broker_str, const std::string &topic_str,
                                                                               const std::string &group_id_str) const;
         std::shared_ptr<kafka_producer_worker> create_producer(const std::string &broker_str, const std::string &topic_str) const;
+        std::shared_ptr<kafka_producer_worker> create_producer(const std::string &bootstrap_server) const;
     };
 
 }
--- a/src/tmx/TmxUtils/src/kafka/kafka_consumer_worker.cpp
+++ b/src/tmx/TmxUtils/src/kafka/kafka_consumer_worker.cpp
@@ -94,6 +94,7 @@ namespace tmx::utils
     void kafka_consumer_worker::stop()
     {
         _run = false;
+        //Close and shutdown the consumer.
         _consumer->close();
         /*Destroy kafka instance*/ // Wait for RdKafka to decommission.
         RdKafka::wait_destroyed(5000);
--- a/src/tmx/TmxUtils/src/kafka/kafka_consumer_worker.cpp
+++ b/src/tmx/TmxUtils/src/kafka/kafka_consumer_worker.cpp
@@ -134,7 +134,7 @@ namespace tmx::utils
         switch (message->err())
         {
         case RdKafka::ERR__TIMED_OUT:
-            FILE_LOG(logWARNING) << _consumer->name() << " consume failed: " <<  message->errstr() << std::endl;
+            FILE_LOG(logDEBUG4) << _consumer->name() << " consume failed: " <<  message->errstr() << std::endl;
             break;
         case RdKafka::ERR_NO_ERROR:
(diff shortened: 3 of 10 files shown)


.devcontainer/docker-compose-vscode.yml
```diff
@@ -72,8 +72,9 @@ services:
       DOCKER_HOST_IP:  127.0.0.1
       KAFKA_ADVERTISED_HOST_NAME:  127.0.0.1
       KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
-      KAFKA_CREATE_TOPICS: "time_sync:1:1"
+      KAFKA_CREATE_TOPICS: "time_sync:1:1,modified_spat:1:1,v2xhub_scheduling_plan_sub:1:1,v2xhub_ssm_sub:1:1"
       KAFKA_LOG_DIRS: "/kafka/kafka-logs"
+      KAFKA_OFFSETS_RETENTION_MINUTES: 1
```
paulbourelly999, line 77: I assume this is a configuration to set how long Kafka will maintain offsets for a consumer group. Is that correct? Can you explain why adding this configuration is important?

src/tmx/TmxUtils/src/kafka/kafka_client.h
```diff
@@ -15,6 +15,7 @@ namespace tmx::utils
         std::shared_ptr<kafka_consumer_worker> create_consumer(const std::string &broker_str, const std::string &topic_str,
                                                                               const std::string &group_id_str) const;
         std::shared_ptr<kafka_producer_worker> create_producer(const std::string &broker_str, const std::string &topic_str) const;
+        std::shared_ptr<kafka_producer_worker> create_producer(const std::string &bootstrap_server) const;
```
paulbourelly999, line 18: This method is to create a producer that is not tied to a topic right?

src/tmx/TmxUtils/src/kafka/kafka_consumer_worker.cpp
```diff
@@ -42,6 +42,12 @@ namespace tmx::utils
             return false;
         }
 
+        if (conf->set(ENABLE_AUTO_COMMIT, "true", errstr) != RdKafka::Conf::CONF_OK)
```
paulbourelly999, line 45: I can look at documentation as well but just curious is you know what additional configuration this is setting?

src/tmx/TmxUtils/src/kafka/kafka_consumer_worker.cpp
```diff
@@ -88,6 +94,7 @@ namespace tmx::utils
     void kafka_consumer_worker::stop()
     {
         _run = false;
+        _consumer->close();
```
paulbourelly999, line 97: Can we leave an inline comment that this close call removes the consumer from the consumer group in the kafka broker.

src/tmx/TmxUtils/src/kafka/kafka_consumer_worker.cpp
```diff
@@ -134,7 +141,7 @@ namespace tmx::utils
         switch (message->err())
         {
         case RdKafka::ERR__TIMED_OUT:
-            FILE_LOG(logWARNING) << _consumer->name() << " consume failed: " <<  message->errstr() << std::endl;
+            FILE_LOG(logDEBUG4) << _consumer->name() << " consume failed: " <<  message->errstr() << std::endl;
```
paulbourelly999, line 144: What do you think about setting this back to log WARNING and just setting the timeout value large to something that indicates an error like 60s or something.

src/tmx/TmxUtils/src/kafka/kafka_producer_worker.cpp
```diff
@@ -33,13 +33,27 @@ namespace tmx::utils
     {
     }
 
+    kafka_producer_worker::kafka_producer_worker(const std::string &brokers)
+        : _broker_str(brokers), _run(true)
+    {
+        init_producer();		
+    }
+
     bool kafka_producer_worker::init()
```
paulbourelly999, line 42: What is the purpose of having two methods here init and init producer and one calls the other. Is it just to reduce method complexity? Also why are we calling the init_producer in the constructor? Seems like both init producer and init_topic are called by init which should be called either way after the constructor. Am I missing something?

src/v2i-hub/CARMAStreetsPlugin/src/CARMAStreetsPlugin.cpp
```diff
@@ -64,96 +67,44 @@ void CARMAStreetsPlugin::UpdateConfigSettings() {
...
-	RdKafka::Conf *tconf_sp = RdKafka::Conf::create(RdKafka::Conf::CONF_TOPIC);
-	RdKafka::Conf *tconf_ssm = RdKafka::Conf::create(RdKafka::Conf::CONF_TOPIC);
-	if(!tconf_spat && !tconf_sp)
+	kafka_client client;
+
+	//Producer
+	_kafka_producer_ptr = client.create_producer(kafkaConnectString);
+
+	//Consumers
+	auto uuid = boost::uuids::random_generator()();
```
paulbourelly999, line 83: We talked about removing this. There is no reason to have a configuration consumer group id if it is appended with random strings.

paulbourelly999 (commented):
Please remove the random hash string appending functionality from the consumer group id.

src/tmx/TmxUtils/src/kafka/kafka_consumer_worker.cpp
dan-du-car: I think it will still print the log constantly when we are in between testing and no data is consumed during the time.

src/tmx/TmxUtils/src/kafka/kafka_producer_worker.cpp
dan-du-car: It is to support creating producer without a topic, we only want to init the producer only. But the original init() func has both init_producer and init_topic portion. So I split them into small chunk which will allow the orignal init() works  as it, and allows for reuse of init_producer in support of creating a producer without a topic.

src/v2i-hub/CARMAStreetsPlugin/src/CARMAStreetsPlugin.cpp
dan-du-car: I think this UUID is the temporary fix to prevent two consumers from joining the same group causing rebalancing.

src/tmx/TmxUtils/src/kafka/kafka_consumer_worker.cpp
dan-du-car: Added

src/v2i-hub/CARMAStreetsPlugin/src/CARMAStreetsPlugin.cpp
paulbourelly999: @dan-du-car can we remove this from the PR description as well so the release notes accurately reflect changes.

src/tmx/TmxUtils/src/kafka/kafka_consumer_worker.cpp
paulbourelly999: I do not think we have any kafka messages that have intervals even close to 60s so it should never print unless we are not receiving kafka messages

dan-du-car: yeah, It could happen as we stop sending BSM/MP/MOM when the vehicle disengage and before the next round of testing. The while loop in the message consumption will keep checking if there are data from those topics.

src/v2i-hub/CARMAStreetsPlugin/src/CARMAStreetsPlugin.cpp
dan-du-car: Done

src/tmx/TmxUtils/src/kafka/kafka_producer_worker.cpp
paulbourelly999: To be honest I would prefer you create a super class of kafka producer that is kafka_topicless_producer or the other way around. Kafka Topic Producer inherits from Kafka producer. I think this is weird because it is unclear which init methods need to be called for each or am I missing something.

paulbourelly999: Maybe we can table that as a future improvement since the kafka_client class abstracts this  but could you at least explain how a producer with a topic would be created vs a producer without a topic to me. Is the producer with a topic required to initialize the topic while the other producer only initializes the producer? Is that correct?

dan-du-car: See the latest code. I removed the init_producer from constructor and allow the plugin to call it and give the plugin a bit more control like the consumers calling init().

src/tmx/TmxUtils/src/kafka/kafka_consumer_worker.cpp
paulbourelly999: Isn't this unique to consumers.

paulbourelly999 (approved):

src/tmx/TmxUtils/src/kafka/kafka_consumer_worker.cpp
dan-du-car: I think so, consumers needs to continuously poll Kafka for data with the while loop. Every poll request, the consumer is telling the Kafka broker that it is alive.

Issue #549, opened 2023-07-13
Move CARMA-Streets Time Synchronization to CARMA-Streets Plugin


paulbourelly999:
### Summary

Currently the CDASim Adapter handles forwarding Time Sync messages directly to CARMA-Streets. A more appropriate implementation of this functionality would be to allow the CARMA-Streets plugin to extend our PluginClientTimeAware class to make it aware of simulation.

### Reasoning for new functionality

This would be a more intuitive split of responsibilities and would allow the CDASimAdapter to be Kafka independent

PR #550, opened 2023-07-14
Issue-549: Moved Kafka time producer from CDASimAdapter to CARMA-Stre…

paulbourelly999:
…ets plugin

+ Update CARMA-Streets Plugin to be simulation time aware

# PR Details
## Description

Move Kafka time producer for CARMA-Streets time synchronization to CARMA-Streets plugin. This removes unnecessary dependency on kafka for CDASimAdapter and also makes CARMA-Streets Plugin contain entire interface to CARMA-Streets including simulation. CARMA-Streets Plugin now extends our simulation time aware PluginClient which can be further extended to listen for simulated messages. 
## Related Issue
#549

## How Has This Been Tested?
Locally integration testing with python time step script to send udp time step messages.

## Types of changes

## Checklist:


src/tmx/TmxUtils/src/PluginClientClockAware.cpp
```diff
@@ -8,11 +8,11 @@ namespace tmx::utils {
...
     {
         // check for simulation mode enabled by environment variable
-        bool simulationMode = sim::is_simulation_mode();
+        _simulation_mode = sim::is_simulation_mode();
 
         using namespace fwha_stol::lib::time;
-        clock = std::make_shared<CarmaClock>(simulationMode);
-        if (simulationMode) {
+        clock = std::make_shared<CarmaClock>(_simulation_mode);
+        if (_simulation_mode) {
```
dan-du-car, line 15: Do we need the if statement to only apply filter when it is in simulation mode? Is it possible for CDAAdapterPlugin to broadcast TimeSyncMessage when it is not in simulation mode?

src/tmx/TmxUtils/src/kafka/kafka_producer_worker.h
```diff
@@ -60,6 +60,19 @@ namespace tmx::utils
              * @param broker_str network address of kafka broker.
              */
             explicit kafka_producer_worker(const std::string &brokers);
+            /**
+             * @brief Destroy the kafka producer worker object. Calls stop on producer to clean up resources.
+             */
+            virtual ~kafka_producer_worker();
+            // Rule of 5 because destructor is define (https://www.codementor.io/@sandesh87/the-rule-of-five-in-c-1pdgpzb04f)
+            // delete copy constructor
+            kafka_producer_worker(kafka_producer_worker& other) = delete;
```
dan-du-car, line 69: What is the purpose of deleting the copy and move constructor?

src/tmx/TmxUtils/src/PluginClientClockAware.cpp
paulbourelly999: No it is not possible and also not valid. TimeSync messages are only valid when we are running in simulation. Otherwise we use machine time.

src/tmx/TmxUtils/src/kafka/kafka_producer_worker.h
paulbourelly999: I left a comment as well

src/tmx/TmxUtils/src/PluginClientClockAware.cpp
paulbourelly999: That is incorrect. Simulation mode will be false in real world deployments. Here the PluginClockAwareClient will not subscribe to TimeSync messages so this is valid in my opinion

paulbourelly999: To clarify, the CDASimAdapter will only be active in simulation mode, which means we will only broadcast TimeSync messages in Simulation Mode. The PluginClockAwareClient is a PluginClient that is able to configurably use Simulation time or real time. It will only subscribe to TimeSync messages when in simulation mode. So on the broadcast side the plugin will not be active to produce time sync messages if not in simulation. On the filter side, even if something else did produce a time sync message, we will not add a filter for it if not in simulation mode.

dan-du-car (approved):

PR #564, opened 2023-11-08
Rsu health monitor

dan-du-car:
# PR Details
## Description
RSU health monitor plugin is a V2xHub plugin that interface with its connected RSU directly via SNMP protocol. The plugin is responsible for monitoring the connected RSU status by constantly ping RSU device. For detailed design of this plugin, refer to https://usdot-carma.atlassian.net/wiki/spaces/WFD2/pages/2640740360/RSU+Health+Monitor+Plugin+Design .

## Related Issue
NA

## Motivation and Context
Data visualization

## How Has This Been Tested?
Integration test

## Types of changes

## Checklist:


src/tmx/Messages/include/RSUStatusMessage.h
```diff
@@ -0,0 +1,25 @@
...
+#include "MessageTypes.h"
+
+
+namespace tmx::messages {
+
+
+class RSUStatusMessage : public tmx::message
+{
+	public:
+		RSUStatusMessage() {}
```
paulbourelly999, line 14: Does this message not have any attributes?

src/tmx/TmxUtils/src/SNMPClient.cpp
```diff
@@ -0,0 +1,323 @@
...
+    // Client defaults to SNMPv3
+    snmp_client::snmp_client(const std::string& ip, const int& port, const std::string& community, 
+        const std::string &snmp_user, const std::string &securityLevel, const std::string &authPassPhrase, int snmp_version, int timeout)
+
+        : ip_(ip), port_(port), community_(community),snmp_version_(snmp_version), timeout_(timeout)
+    {
+        
+        PLOG(logDEBUG1) << "Starting SNMP Client";
+        PLOG(logDEBUG1) << "Target device IP address: " << ip_;
+        PLOG(logINFO) << "Target device NTCIP port: " << port_;
```
paulbourelly999, line 14: This logging does not may sense to me. NTCIP port is probably incorrect here. Also we probably want all these log statements to share a log level.

src/v2i-hub/RSUHealthMonitorPlugin/manifest.json
```diff
@@ -0,0 +1,46 @@
...
+		{
+            "key":"RSUOIDConfigMap",
+            "default":"{\"RSUOIDConfig\":[{\"RsuField\":\"rsuGpsOutputString\",\"OID\":\"iso.0.15628.4.1.8.5.0\",\"Required\":false},{\"RsuField\":\"rsuID\",\"OID\":\"iso.0.15628.4.1.17.4.0\",\"Required\":false},{\"RsuField\":\"rsuMibVersion\",\"OID\":\"iso.0.15628.4.1.17.1.0\",\"Required\":false},{\"RsuField\":\"rsuFirmwareVersion\",\"OID\":\"iso.0.15628.4.1.17.2.0 \",\"Required\":false},{\"RsuField\":\"rsuManufacturer\",\"OID\":\"iso.0.15628.4.1.17.5.0\",\"Required\":false},{\"RsuField\":\"rsuIFMIndex\",\"OID\":\"iso.3.6.1.2.1.1.7.0\",\"Required\":false},{\"RsuField\":\"rsuIFMPsid\",\"OID\":\"iso.3.6.1.2.1.1.7.0\",\"Required\":false},{\"RsuField\":\"rsuIFMDsrcMsgId\",\"OID\":\"iso.3.6.1.2.1.1.7.0\",\"Required\":false},{\"RsuField\":\"rsuIFMTxMode\",\"OID\":\"iso.3.6.1.2.1.1.7.0\",\"Required\":false},{\"RsuField\":\"rsuIFMTxChannel\",\"OID\":\"iso.3.6.1.2.1.1.7.0\",\"Required\":false},{\"RsuField\":\"rsuIFMEnable\",\"OID\":\"iso.3.6.1.2.1.1.7.0\",\"Required\":false},{\"RsuField\":\"rsuIFMStatus\",\"OID\":\"iso.3.6.1.2.1.1.7.0\",\"Required\":false},{\"RsuField\":\"rsuMode\",\"OID\":\"iso.0.15628.4.1.99.0\",\"Required\":false}]}",
```
paulbourelly999, line 42: This does not make sense to me. Aren't these OIDs standardized. Maybe it would be better to specify a document version here and then have different headers that contain the actual OIDs.

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorPlugin.cpp
```diff
@@ -0,0 +1,220 @@
...
+                        Json::FastWriter fasterWirter;
+                        string json_str = fasterWirter.write(rsuStatusJson);
+                        tmx::messages::RSUStatusMessage sendRsuStatusMsg;
+                        sendRsuStatusMsg.set_contents(json_str);
+                        BroadcastMessage(sendRsuStatusMsg);
+                        PLOG(logINFO) << "Broadcast RSU status:  " << json_str;
+                    }
+                }
+            }
+            catch (std::exception &ex)
```
paulbourelly999, line 102: This should be const

src/tmx/Messages/include/RSUStatusMessage.h
dan-du-car: No, we just need the content which is from tmx::message.

src/tmx/TmxUtils/src/SNMPClient.cpp
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/564/commits/c0bac101e0cef01c154929024be6ab01b263785b

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorPlugin.cpp
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/564/commits/c0bac101e0cef01c154929024be6ab01b263785b

src/tmx/TmxUtils/src/RSU_MIB_4_1.h
```diff
@@ -0,0 +1,69 @@
+#pragma once
+namespace tmx
```
paulbourelly999, line 2: You can do something like this to make this less verbose
```
namespace tmx::utils::rsu41::mid::oid
```

src/v2i-hub/RSUHealthMonitorPlugin/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+#############
+enable_testing()
+include_directories(${PROJECT_SOURCE_DIR}/src)
+add_library(${PROJECT_NAME}_lib src/RSUHealthMonitorWorker.cpp)
+target_link_libraries(${PROJECT_NAME}_lib PUBLIC ${TMXAPI_LIBRARIES} 
+                                        ${ASN_J2735_LIBRARIES} 
+                                        ${MYSQL_LIBRARIES} 
+                                        ${MYSQLCPPCONN_LIBRARIES}
+                                        tmxutils
+                                        NemaTode
```
paulbourelly999, line 26: What is nematode?

src/v2i-hub/RSUHealthMonitorPlugin/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+                                        ${UUID_LIBRARY})
+set(BINARY ${PROJECT_NAME}_test)
+file(GLOB_RECURSE TEST_SOURCES LIST_DIRECTORIES false test/*.h test/*.cpp)
+set(SOURCES ${TEST_SOURCES} WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/test)
+add_executable(${BINARY} ${TEST_SOURCES})
+add_test(NAME ${BINARY} COMMAND ${BINARY})
+target_link_libraries(${BINARY} PUBLIC ${PROJECT_NAME}_lib gtest NemaTode)
```
paulbourelly999, line 33: you do not need to link NemaTode here if you made it a public depedency of you lib and you are including you lib

src/v2i-hub/RSUHealthMonitorPlugin/manifest.json
```diff
@@ -0,0 +1,51 @@
+{
+    "name": "RSUHealthMonitor",
+    "description": "Monitor RSU health status",
+    "version": "@PROJECT_VERSION@",
+    "exeLocation": "/bin/RSUHealthMonitorPlugin",
+    "coreIpAddr":"127.0.0.1",
+	"corePort":24601,
+    "messageTypes": [],
```
paulbourelly999, line 8: You should probably add the RSU Health message to message types here.

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorPlugin.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+    void RSUHealthMonitorPlugin::OnConfigChanged(const char *key, const char *value)
+    {
+        PluginClient::OnConfigChanged(key, value);
+        UpdateConfigSettings();
+    }
+
+    void RSUHealthMonitorPlugin::PeriodicRSUStatusReq()
+    {
+        while (true)
```
paulbourelly999, line 52: Not a fan of non exitable loops. Consider using one of the built in tmx wrappers on thread instead of thread directly like `ThreadTimer` or `ThreadWorker`. Even if you continue to use thread directly there should probably be some sort of exit condition. If you extend thread working, you can implement the DoWork function and it makes this logic unit testable since it will no longer be inside you plugin class.

src/v2i-hub/RSUHealthMonitorPlugin/CMakeLists.txt
dan-du-car: https://github.com/ckgt/NemaTode We use this to parse the RSU GPS Nmea string.

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorPlugin.cpp
```diff
@@ -0,0 +1,160 @@
...
+            }
+            catch (const std::exception &ex)
+            {
+                PLOG(logERROR) << ex.what();
+            }
+            this_thread::sleep_for(chrono::seconds(_interval));
+        }
+    }
+
+    Json::Value RSUHealthMonitorPlugin::getRSUStatus()
```
paulbourelly999, line 87: Is there are reason this is not in the RSU Worker. Here it is not unit testable?

src/v2i-hub/RSUHealthMonitorPlugin/manifest.json
dan-du-car: What is the purpose of adding this message types here?

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorPlugin.cpp
```diff
@@ -0,0 +1,160 @@
...
+    {
+        if (!_rsuWorker)
+        {
+            PLOG(logERROR) << "RSU status update call failed due to fail to initialize RSU worker!";
+            return Json::nullValue;
```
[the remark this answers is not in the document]
dan-du-car: The caller checks null value to determine the action rather than requiring all callers to catch them later.

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorPlugin.cpp
dan-du-car: It is sending the SNMP request which requires the connection to the RSU, and also require the global RSU Status config variable from the plugin.

paulbourelly999 (changes requested):
Please check out the sonar scan code smells. A lot of these seem pretty easy to address. Also you need to update the sonar.properties file to include your new plugin

dan-du-car:
> Please check out the sonar scan code smells. A lot of these seem pretty easy to address. Also you need to update the sonar.properties file to include your new plugin

Added https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/564/commits/5f7858e46c999a341f91ed0f8685ca8b14711552

src/v2i-hub/RSUHealthMonitorPlugin/CMakeLists.txt
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/564/commits/5f7858e46c999a341f91ed0f8685ca8b14711552

src/tmx/TmxUtils/src/RSU_MIB_4_1.h
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/564/commits/d46893578a0d09dfdf8decc8166c102da3e3a449

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorPlugin.cpp
dan-du-car: Is there any issue with retrieving the RSU info and broadcast immediately? I am trying to avoid the complexity of two threads accessing the same global variable where the RSU status is stored.

dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/564/commits/7a4238acb369a0a8baa26b437c9b0ab59e48960a

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorPlugin.cpp
paulbourelly999: Both the SNMP Client and any configuration can be passed to another class to make this call right? Just my opinion, but when creating new V2X-Hub Plugins since we can not unit test the code there, we should be moving most of it to other cpp files. Also for maintainability reasons we should not couple the logic for making snmp request, json serialization and the plugin core logic together if we do not need to.

src/v2i-hub/RSUHealthMonitorPlugin/manifest.json
paulbourelly999: I guess in this case it won't work but for plugins that receive a certain message, it adds a status item that includes a count for the number of that type of message received by the plugin.

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorPlugin.cpp
```diff
@@ -0,0 +1,97 @@
...
+        // Send SNMP call to RSU periodically at configurable interval.
+        _timerThId = _rsuStatusTimer->AddPeriodicTick([this]()
+                                                      {
+            // Periodic SNMP call to get RSU status based on RSU MIB version 4.1
+            auto rsuStatusJson =  _rsuWorker->getRSUStatus(_rsuMibVersion, _rsuIp, _snmpPort, _securityUser, _authPassPhrase, _securityLevel, SEC_TO_MICRO);
+            PLOG(logINFO) << "Updating _interval: " << _interval;
+            //Broadcast RSU status periodically at _interval
+            BroadcastRSUStatus(rsuStatusJson); },
+                                                      std::chrono::milliseconds(_interval * SEC_TO_MILLI));
+        _rsuStatusTimer->Start();
```
paulbourelly999, line 24: Can we move this start to when the Plugin is enabled on config changes.

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorWorker.cpp
```diff
@@ -0,0 +1,175 @@
...
+            auto latitude_str = ss.str();
+            std::stringstream sss;
+            sss << std::setprecision(8) << std::fixed << gps.fix.longitude << std::endl;
+            auto longitude_str = sss.str();
+            result.insert({std::stod(latitude_str), std::stod(longitude_str)});
+            PLOG(logDEBUG) << "Parse GPS NMEA string: " << gps_nmea_data << ". Result (Latitude, Longitude): (" << latitude_str << "," << longitude_str << ")";
+        }
+        catch (nmea::NMEAParseError &e)
+        {
+            fprintf(stderr, "Error:%s\n", e.message.c_str());
```
paulbourelly999, line 112: Why are we using fprintf instead of PLOG

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorWorker.cpp
```diff
@@ -0,0 +1,175 @@
...
+            std::stringstream ss;
+            ss << std::setprecision(8) << std::fixed << gps.fix.latitude << std::endl;
+            auto latitude_str = ss.str();
+            std::stringstream sss;
+            sss << std::setprecision(8) << std::fixed << gps.fix.longitude << std::endl;
+            auto longitude_str = sss.str();
+            result.insert({std::stod(latitude_str), std::stod(longitude_str)});
+            PLOG(logDEBUG) << "Parse GPS NMEA string: " << gps_nmea_data << ". Result (Latitude, Longitude): (" << latitude_str << "," << longitude_str << ")";
+        }
+        catch (nmea::NMEAParseError &e)
```
paulbourelly999, line 110: Should be constant

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorWorker.cpp
```diff
@@ -0,0 +1,175 @@
...
+        return result;
+    }
+
+    Json::Value RSUHealthMonitorWorker::getRSUStatus(const RSUMibVersion &mibVersion, const string &_rsuIp, uint16_t &_snmpPort, const string &_securityUser, const string &_authPassPhrase, const string &_securityLevel, long timeout)
+    {
+        auto rsuStatusConfigTbl = GetRSUStatusConfig(mibVersion);
+        if (rsuStatusConfigTbl.size() == 0)
+        {
+            PLOG(logERROR) << "RSU status update call failed due to the RSU status config table is empty!";
+            return Json::nullValue;
```
paulbourelly999, line 123: Why are we printing error and returning null instead of throwing an exception

src/tmx/TmxUtils/src/SNMPClient.cpp
```diff
@@ -0,0 +1,306 @@
...
+        return true;
+    }
+
+    // Backup GET function for use with prexisting ERVCloudForwarding client
+    std::string snmp_client::SNMPGet(const std::string &req_oid)
+    {
+        snmp_pdu *response;
+
+        std::string result = "";
+        auto pdu = snmp_pdu_create(SNMP_MSG_GET);
```
paulbourelly999, line 241: There is a member variable with the same name and overlapping scope. We should probably reconsider naming here

src/tmx/TmxUtils/CMakeLists.txt
```diff
@@ -1,7 +1,9 @@
 PROJECT ( tmxutils CXX )
 FILE (GLOB_RECURSE HEADERS "src/" "*.h*")
 FILE (GLOB_RECURSE SOURCES  "src/" "*.c*")
-
```
paulbourelly999, line 4: This is optional but I think it would be helpful if we separate some of this into sub packages like `TmxUtils/src/snmp_client` similar to what we have done with kafka.

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorWorker.h
```diff
@@ -0,0 +1,96 @@
...
+
+    struct RSUFieldOIDStruct
+    {
+        string field;
+        string oid;
+        bool required; // Indicate whether this field is required to before broadcasting the RSU status.
+    };
+
+    /**
+     * RSUStatusTable is custome defined RSU status information.
```
paulbourelly999, line 35: *custom

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorWorker.h
```diff
@@ -0,0 +1,96 @@
...
+     * The fields are the subset of fields from RSU MIB definition https://github.com/certificationoperatingcouncil/COC_TestSpecs/blob/master/AppNotes/RSU/RSU-MIB.txt
```
paulbourelly999, line 36: The fields are a subset of the fields from the RSU MIB definition used to quantify the health of the RSU.

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorWorker.h
```diff
@@ -0,0 +1,96 @@
...
+     */
+    using RSUStatusConfigTable = vector<RSUFieldOIDStruct>;
+
+    class RSUHealthMonitorWorker
+    {
+    private:
+        // A map of RSU MIB version used and RSUStatusTable
+        shared_ptr<map<RSUMibVersion, RSUStatusConfigTable>> _RSUSTATUSConfigMapPtr;
```
paulbourelly999, line 44: Why do we have a map between mib version and RSU Status Config table? Will we populate this map with all versions regardless of the configured version and then just use the configured version?

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorWorker.h
```diff
@@ -0,0 +1,96 @@
...
+        RSUStatusConfigTable GetRSUStatusConfig(const RSUMibVersion &mibVersion) const;
+
+        /**
+         * @brief determine if all required fields in the RSU config map _RSUSTATUSConfigMapPtr present in the input fields
+         * Use _RSUSTATUSConfigMapPtr RSU status config map that defines all fields and whether the fields are required.
+         * @param RSUMibVersion RSU MIB version
+         * @param vector<string> Input RSU fields to verify
+         * @return True if all required fields found. Otherwise, false.
+         */
+        bool isAllRequiredFieldsPresent(const RSUMibVersion &mibVersion, const vector<string> &fields) const;
```
paulbourelly999, line 68: Can we rename this method to validate. Also I am not sure I understand this method. Is it meant to validate a RSU status message or is it meant to confirm all the required OIDs are obtainable from SNMP client?

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorWorker.h
```diff
@@ -0,0 +1,96 @@
...
+
+        /**
+         * @brief Parse NMEA GPS sentense and return GPS related data
```
paulbourelly999, line 71: *Sentence Although maybe we want to reword this to be Parse NMEA GPS String ...
Also what is GPS related data? Is this just a long lat coordinate. Why do we need a map if its just a long lat coordinate?

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorWorker.cpp
```diff
@@ -0,0 +1,175 @@
...
+            RSUFieldOIDStruct rsuMode = {"rsuMode", RSU_MODE, true};
+            rsuStatusTbl.push_back(rsuMode);
+
+            RSUFieldOIDStruct rsuChanStatus = {"rsuChanStatus", RSU_CHAN_STATUS, true};
+            rsuStatusTbl.push_back(rsuChanStatus);
+        }
+        return rsuStatusTbl;
+    }
+
+    bool RSUHealthMonitorWorker::isAllRequiredFieldsPresent(const RSUMibVersion &mibVersion, const vector<string> &fields) const
```
paulbourelly999, line 65: If this is a validation method I suggest for clarity sake you pass in the config directly instead of passing in the mib version and calling Get config. I think it is cleaner for the validate method to take parameters ( message to validate, list of required fields)

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorPlugin.cpp
```diff
@@ -0,0 +1,97 @@
...
+    {
+        _rsuWorker = std::make_shared<RSUHealthMonitorWorker>();
+        _rsuStatusTimer = make_unique<ThreadTimer>();
+        UpdateConfigSettings();
+
+        // Send SNMP call to RSU periodically at configurable interval.
+        _timerThId = _rsuStatusTimer->AddPeriodicTick([this]()
+                                                      {
+            // Periodic SNMP call to get RSU status based on RSU MIB version 4.1
+            auto rsuStatusJson =  _rsuWorker->getRSUStatus(_rsuMibVersion, _rsuIp, _snmpPort, _securityUser, _authPassPhrase, _securityLevel, SEC_TO_MICRO);
```
paulbourelly999, line 19: This is confusing to me. So you get the RSU status here by querying the SNMP server and you return a JSON. Then you wait until the broadcast method to check that the required information is present? You do not have to change this, it just seems weird to me. I would imagine that the method that gets the RSU status either directly turns it into the tmx message (including validation) or the next method turns it into a tmx message.

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorWorker.cpp
dan-du-car: We encourage users to update the RSU config table by updating the mib version with the supported version, but it is not a fatal program issue.

src/tmx/TmxUtils/CMakeLists.txt
dan-du-car: I would leave this to enhancement later if we would like to reorganize all the files.

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorWorker.h
dan-du-car: yes, we will populate the map with all versions if exist like NTCIP1218 and RSU4.1. But the users can choose which version to use by specifying the mib version through the UI?

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorWorker.h
dan-du-car: sure, update to validate. Yes, it is to make sure we obtain all the required fields.

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorWorker.h
dan-du-car: Yes, what data structure do you suggest if not map?

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorPlugin.cpp
dan-du-car: If the JSON is empty due to some SNMP timeout or other issue, we stop any further logic like creating TMX message, broadcasting message.

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorWorker.cpp
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/564/commits/a49366ea8030afa8c668ec5879cf8e2d5cf69065

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorWorker.h
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/564/commits/a49366ea8030afa8c668ec5879cf8e2d5cf69065

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorWorker.h
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/564/commits/a49366ea8030afa8c668ec5879cf8e2d5cf69065

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorWorker.h
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/564/commits/a49366ea8030afa8c668ec5879cf8e2d5cf69065

src/tmx/TmxUtils/src/SNMPClient.cpp
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/564/commits/a49366ea8030afa8c668ec5879cf8e2d5cf69065

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorWorker.cpp
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/564/commits/a49366ea8030afa8c668ec5879cf8e2d5cf69065

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorWorker.cpp
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/564/commits/a49366ea8030afa8c668ec5879cf8e2d5cf69065

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorPlugin.cpp
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/564/commits/a49366ea8030afa8c668ec5879cf8e2d5cf69065

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorWorker.cpp
paulbourelly999: Exceptions do not need to be fatal but this is fine for now.

paulbourelly999 (approved):

PR #565, opened 2023-11-17
v2x hub Integration: Implement telematic plugin to subscribe to all TMX messages from V2xHub

dan-du-car:
# PR Details
## Description
In order to publish data from V2xHub to cloud a telematics bridge is needed. This story is the implementation story for the telematic module to collect J2735 data from V2xHub. The design refers to https://usdot-carma.atlassian.net/wiki/spaces/WFD2/pages/2640379977/V2xHub+WFD+Bridge+Design 

Connect to V2xHub

Subscribe all TMX messages from V2xhub.

If there are J2735 message payloads in the TMX messages, decode the J2735 into human readable message.

Convert TMX messages into messages in JSON format.

## Related Issue
NA

## Motivation and Context
CDA telematic

## How Has This Been Tested?
Unit test
Integration test

## Types of changes

## Checklist:


src/v2i-hub/SpatPlugin/test/testPedestrianDetectionForSPAT.cpp
```diff
@@ -73,7 +73,7 @@ TEST(PedestrianDetectionForSPAT, updateEncodedSpat)
         EXPECT_EQ(spatPtr->intersections.list.array[0]->states.list.array[0]->signalGroup, 1);
         EXPECT_EQ(spatPtr->intersections.list.array[0]->states.list.array[1]->signalGroup, 2);
         ASSERT_NE(spatPtr->intersections.list.array[0]->maneuverAssistList, nullptr);
-        EXPECT_EQ(spatPtr->intersections.list.array[0]->maneuverAssistList->list.count, 1);
+        EXPECT_NE(spatPtr->intersections.list.array[0]->maneuverAssistList->list.count, 0);
```
paulbourelly999, line 76: Why this change?

dan-du-car: This unit test sometimes failed. https://github.com/usdot-fhwa-OPS/V2X-Hub/actions/runs/6895436333/job/18759375988

dan-du-car: I ran the unit test in local and the list count sometimes equals to 1, sometimes equals to 2.

ext/build.sh
```diff
@@ -55,4 +55,14 @@ cd NemaTode
...
\ No newline at end of file
+popd
+
+# Nats C API
+pushd /tmp
+git clone https://github.com/nats-io/nats.c
+cd nats.c
+cmake . -DNATS_BUILD_NO_SPIN=ON
+make -j${numCPU}
+make install
```
paulbourelly999, line 66: After we install this dependency, we should delete the source code.

src/v2i-hub/SpatPlugin/test/testPedestrianDetectionForSPAT.cpp
paulbourelly999: This is a issue. Can we create a bug for this is we do not plan to address it in this PR and leave a TODO comment here outlining what you found.

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorWorker.cpp
```diff
@@ -145,7 +145,11 @@ namespace RSUHealthMonitor
                     }
                     else if (success)
                     {
-                        rsuStatuJson.append(populateJson(config.field, responseVal));
+                        auto json = populateJson(config.field, responseVal);
```
paulbourelly999, line 148: What changed here for the RSU Health Monitor Plugin

src/v2i-hub/TelematicBridgePlugin/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+PROJECT (TelematicBridgePlugin VERSION 7.5.1 LANGUAGES CXX)
+
+set (TMX_PLUGIN_NAME "Telematic Bridge")
+
+find_package(RapidJSON REQUIRED)
+
+BuildTmxPlugin()
+TARGET_LINK_LIBRARIES ( ${PROJECT_NAME} tmxutils jsoncpp)
```
paulbourelly999, line 8: if we are not linking rapidjson, what is the point of finding the package?

src/v2i-hub/TelematicBridgePlugin/src/TelematicBridgeMsgWorker.cpp
```diff
@@ -0,0 +1,184 @@
+#include "TelematicBridgeMsgWorker.h"
+
+namespace TelematicBridge
+{
+    bool TelematicBridgeMsgWorker::HexToBytes(const string &hexPaylod, vector<char> &byteBuffer)
```
paulbourelly999, line 5: What is this method for?

src/v2i-hub/TelematicBridgePlugin/src/TelematicBridgeMsgWorker.cpp
```diff
@@ -0,0 +1,184 @@
...
+            nullptr,
+            ATS_UNALIGNED_BASIC_PER,
+            &asn_DEF_MessageFrame,
+            (void **)&messageFrame,
+            byte_buffer.data(),
+            byte_buffer.size());
+
+        if (decode_rval.code != RC_OK)
+        {
+            erroross.str("");
```
paulbourelly999, line 65: What is this erroross object?

src/v2i-hub/TelematicBridgePlugin/src/TelematicBridgeMsgWorker.h
```diff
@@ -0,0 +1,68 @@
...
+namespace TelematicBridge
+{
+    using buffer_structure_t = struct buffer_structure
+    {
+        char *buffer;          // buffer array
+        size_t buffer_size;    // this is really where we will write next.
+        size_t allocated_size; // this is the total size of the buffer.
+    };
+
+    class TelematicBridgeMsgWorker
```
paulbourelly999, line 21: Does this object have any state information? Is there a reason we need a class for this or could we just define functions here without a class

src/v2i-hub/TelematicBridgePlugin/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+####################################################
+##################  Testing  #######################
+####################################################
+enable_testing()
+include_directories(${PROJECT_SOURCE_DIR}/src)
+add_library(${PROJECT_NAME}_lib src/TelematicBridgeMsgWorker.cpp)
+target_link_libraries(${PROJECT_NAME}_lib  PUBLIC tmxutils jsoncpp)
+file(GLOB_RECURSE TEST_SOURCES LIST_DIRECTORIES false test/*.h test/*.cpp)
+add_executable(${PROJECT_NAME}_test ${TEST_SOURCES})
+target_link_libraries(${PROJECT_NAME}_test PRIVATE ${PROJECT_NAME}_lib gtest)
```
paulbourelly999, line 19: If you link the following you do not need to create a `main.cpp` file for testing.
```cmake
target_link_libraries(${PROJECT_NAME}_test PRIVATE ${PROJECT_NAME}_lib GTest::Main
```

src/v2i-hub/TelematicBridgePlugin/manifest.json
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "TelematicBridge",
+    "description": "Plugin that listens for TMX messages and forward them to the Telematic cloud services.",
+    "version": "@PROJECT_VERSION@",
+    "exeLocation": "/bin/TelematicBridgePlugin",
+    "coreIpAddr": "127.0.0.1",
+    "corePort": 24601,
+    "messageTypes": [],
```
paulbourelly999, line 8: Update Message Types here if we want status counts for the messages this plugin handles.

src/v2i-hub/TelematicBridgePlugin/src/TelematicBridgePlugin.cpp
```diff
@@ -0,0 +1,39 @@
+#include "TelematicBridgePlugin.h"
+
+namespace TelematicBridge
+{
+    TelematicBridgePlugin::TelematicBridgePlugin(const string &name) : PluginClient(name)
+    {
+        AddMessageFilter("*", "*", IvpMsgFlags_None);
```
paulbourelly999, line 7: Will this only pickup messages with no IvpMsgFlags? I imagine this filter should look essentially identical to the ImmediateForward filter, except it should not require a DSRC flag. 

The immediate forward filer looks like this :
```cpp
AddMessageFilter("J2735", "*", IvpMsgFlags_RouteDSRC);
```

src/v2i-hub/SpatPlugin/test/testPedestrianDetectionForSPAT.cpp
dan-du-car: https://github.com/usdot-fhwa-OPS/V2X-Hub/issues/566

scripts/install_dependencies.sh
```diff
@@ -31,6 +31,9 @@ DEPENDENCIES="build-essential \
     wget \
     zip \
     zlib1g \
+    rapidjson-dev \
```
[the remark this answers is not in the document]
dan-du-car: I think the xml to json library use the rapidjson lib

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorWorker.cpp
dan-du-car: The JSON was an array. My intention is to create a JSON object.

paulbourelly999 (changes requested):
Although I think it is unnecessary to pull in a new JSON dependency for the Telematics bridge when we already have jsoncpp included, I think it probably not worth the effort now to try and change the logic for the existing code. Some changes that I think are worth while:

Removing any unnecessary classes. If something is not stateful, meaning it does not have a state associated with it, I think we can just have free functions instead of creating an object and having static methods associated with it. 

Additionally, I  think its worth while reworking the configuration surrounding the messages we want to forward to telematics. I think this should be very similar to how Immediate Forward works in that we should be able to configure what J2735 messages, as well as what TMX messages get forwarded to Telematics.

Lastly, I am not sure I understand why the processing of J2735 messages for this plugin requires so many custom methods, like converting hex string to bytes (etc). I am unsure how different this plugin is from Message Receiver and Immediate forward. Why do we need these methods there, but not here. Considering using different Handle methods. Maybe instead of ivp message we should be dealing with routeable message.

@paulbourelly999 This plugin is very similar to Immediate Forward plugin. The only difference from immediate Forward is that it decode the J2735 and also subscribe to any other TMX messages other than J2735.

src/v2i-hub/TelematicBridgePlugin/src/TelematicBridgePlugin.cpp
dan-du-car: How do we map the PSID to each TMX message? eg: time sync TMX message, do we have a PSID for that? Which messages use the IvpMsgFlags_RouteDSRC flag? When I test the J2735 message with Message receiver plugin, I can subcribe to the J2735 without IvpMsgFlags_RouteDSRC flag. The intention for this plugin is to discover all messages. So that the plugin can tell telematic system what topics/message available in V2xHub. V2xHub UI does not control what to subscribe to. Instead, the telematic UI will receive a list of all available topics/messages  and decide which messages to send to cloud. Note: we will create a mapping between messages and topics as  the plugin discover messages with the AddMessageFilter func.

dan-du-car: Immediate Forward plugin subscribe to all J2735 message and use IVPMessage rather than routable message.

src/v2i-hub/TelematicBridgePlugin/CMakeLists.txt
dan-du-car: image

src/v2i-hub/TelematicBridgePlugin/src/TelematicBridgeMsgWorker.h
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/565/commits/5ca81e0164304caaa46367ebb61763ded5c7607b

src/v2i-hub/TelematicBridgePlugin/src/TelematicBridgeMsgWorker.cpp
dan-du-car: ostringstream

src/v2i-hub/TelematicBridgePlugin/CMakeLists.txt
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/565/files#diff-44401677ab8e59119c4870aabc314fba94c3a4cb82288919cf11b36257a5efe0

ext/build.sh
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/565/commits/a3b68aa8789a5a9d5e17bd3807628f4f76026d08

paulbourelly999 (approved):

PR #567, opened 2023-11-22
Telematic bridge nats

dan-du-car:
# PR Details
## Description
This is the implementation story for the telematic bridge to stream J2735 data from V2xHub into the cloud  at near real time. The module design refers to https://usdot-carma.atlassian.net/wiki/spaces/WFD2/pages/2640379977/V2xHub+WFD+Bridge+Design The following functionalities should be implemented for this module:

Able to forward the JSON message to telematic cloud when the message type is requested by end users through telematic UI.

Able to register itself with WFD cloud services.

Able to map each TMX message type to each individual topic.  

Able to provide available topics to telematic cloud.

Able to stream the requested TMX message in JSON format to telematic cloud.

## Related Issue
NA

## Motivation and Context
Telematic tool data collection

## How Has This Been Tested?
Unit test and local integration testing

## Types of changes

## Checklist:


paulbourelly999:
@dan-du-car  in the future lets include README documentation about any plugin in Github. If we can move some of the appropriate information from the design documentation on confluence to a README file here that would be great. If not then we can just prioritize this moving forward.

src/v2i-hub/TelematicBridgePlugin/manifest.json
```diff
@@ -11,6 +11,31 @@
             "key": "LogLevel",
             "default": "INFO",
             "description": "The log level for this plugin"
+        },
+        {
+            "key": "NATSUrl",
+            "default": "nats://127.0.0.1:4222",
+            "description": "The log level for this plugin"
```
paulbourelly999, line 18: Incorrect descriptions for all configuration parameters

src/v2i-hub/TelematicBridgePlugin/manifest.json
```diff
@@ -11,6 +11,31 @@
...
             "description": "The log level for this plugin"
+        },
+        {
+            "key": "NATSUrl",
+            "default": "nats://127.0.0.1:4222",
+            "description": "The log level for this plugin"            
+        },
+        {
+            "key": "UnitId",
+            "default": "v2xhub_id",
```
paulbourelly999, line 22: We are currently, in CDASim pulling and infrastructure ID from an environment variable. You may want to do the same here. See CDASimAdapter

src/v2i-hub/TelematicBridgePlugin/manifest.json
```diff
@@ -11,6 +11,31 @@
...
+            "description": "The log level for this plugin"            
+        },
+        {
+            "key": "UnitName",
+            "default": "East Intersection",
+            "description": "The log level for this plugin"            
+        },
+        {
+            "key": "UnitType",
+            "default": "Infrastructure",
```
paulbourelly999, line 32: Not sure about this field.

src/v2i-hub/TelematicBridgePlugin/manifest.json
```diff
@@ -11,6 +11,31 @@
...
+            "description": "The log level for this plugin"            
+        },
+        {
+            "key": "TopicExclusionList",
```
paulbourelly999, line 36: These are not topics but messages

src/v2i-hub/TelematicBridgePlugin/readme.md
```diff
@@ -0,0 +1,94 @@
+## NATS Publisher/Subscriber
```
paulbourelly999, line 1: Can we make this file called README.md instead of readme.md

src/v2i-hub/TelematicBridgePlugin/src/TelematicBridgePlugin.cpp
```diff
@@ -4,6 +4,8 @@ namespace TelematicBridge
 {
     TelematicBridgePlugin::TelematicBridgePlugin(const string &name) : PluginClient(name)
     {
+        _telematicUnitPtr = make_shared<TelematicUnit>();
```
paulbourelly999, line 7: Can this be a unique pt?

src/v2i-hub/TelematicBridgePlugin/src/TelematicUnit.h
```diff
@@ -0,0 +1,218 @@
...
+        static CONSTEXPR const char *EVENT_NAME = "event_name";               // event_name key used to find event_name value from JSON
+        static CONSTEXPR const char *UNIT_ID = "unit_id";                     // unit_id key used to find unit_id value from JSON
+        static CONSTEXPR const char *UNIT_NAME = "unit_name";                 // unit_name key used to find unit_name value from JSON
+        static CONSTEXPR const char *UNIT_TYPE = "unit_type";                 // unit_type key used to find unit_type value from JSON
+        static CONSTEXPR const char *TOPIC_NAME = "topic_name";               // topic_name key used to find topic_name value from JSON
+        static CONSTEXPR const char *TIMESTAMP = "timestamp";                 // timestamp key used to find timestamp value from JSON
+        static CONSTEXPR const char *PAYLOAD = "payload";                     // payload key used to find payload value from JSON
+        static CONSTEXPR const char *TOPICS = "topics";                       // topics key used to find topics value from JSON
```
paulbourelly999, line 57: These are not topics, I would refrain from using this language unless telematics requires it.

dan-du-car: I can change the variable name "TOPICS". We do need the value topics to find JSON value that match this topics key.

src/v2i-hub/TelematicBridgePlugin/src/TelematicBridgePlugin.cpp
dan-du-car: Is there a reason for this change?

src/v2i-hub/TelematicBridgePlugin/manifest.json
dan-du-car: The topic name consist of type_subtype_source. We create a topic for each message because v2xhub does not have a topic.

src/v2i-hub/TelematicBridgePlugin/manifest.json
dan-du-car: This is not the same concept as the infrastructure id. It is a unique identifier we give to a telematic unit to distinguish it from other units like streets and vehicle. Is there any rule when we use the UI configuration parameter and when to use the environment variables? All these parameters make sense when they are together.

src/v2i-hub/TelematicBridgePlugin/manifest.json
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/567/commits/e8678760c97cbb7d5bd212ac07f490793d7d8c63

src/v2i-hub/TelematicBridgePlugin/REAADME.md
```diff
@@ -0,0 +1,94 @@
+## NATS Publisher/Subscriber
+### NATS Connections and registration
```
paulbourelly999, line 2: REAME file is misspelled

src/v2i-hub/TelematicBridgePlugin/REAADME.md
```diff
@@ -0,0 +1,94 @@
...
+#### Telematic plugin sends registration request to telematic server
```
paulbourelly999, line 3: Please update to include an overall plugin description

src/v2i-hub/TelematicBridgePlugin/manifest.json
paulbourelly999: Hmm, so part of the UC for this plugin is to disable health of different RSUs, and therefore infrastructure instances. Coming up with and ID that is completely different from the J2735 intersection ID and our infrastructure ID I think is bad practice. Then we will have 3 different IDs for the same instance, all which are completely indepedent.

src/v2i-hub/TelematicBridgePlugin/manifest.json
paulbourelly999: UnitType does not make sense as a configurable parameter if it is sent to infrastructre. Every V2X-Hub falls under infrastructure. Maybe if we have types like intersection or highway it would make sense but like this is does not make sense.

src/v2i-hub/TelematicBridgePlugin/manifest.json
paulbourelly999: So topics is not a familiar term for users of V2X-HUb. If we are making a connection between topics and message in telematics, I would recommend we use "message" to refer to the messages we are listening to in V2X-Hub and "topic" as the term of the topics we will create on telematics.

I suggest we call this messages here and the term topics is only used to send information to telematics about topics to create and in the telematics UI/Software.

src/v2i-hub/TelematicBridgePlugin/src/TelematicBridgePlugin.h
```diff
@@ -3,21 +3,30 @@
 
 #include "PluginClient.h"
 #include "TelematicBridgeMsgWorker.h"
+#include "TelematicUnit.h"
 
 using namespace tmx::utils;
```
paulbourelly999, line 8: Using namespace is bad practice especially in header files.

src/v2i-hub/TelematicBridgePlugin/src/TelematicUnit.cpp
```diff
@@ -0,0 +1,314 @@
...
+                auto replyStr = natsMsg_GetData(reply);
+                PLOG(logINFO) << "Received registered reply: " << replyStr;
+
+                // Unit is registered when server responds with event information (location, testing_type, event_name)
+                isRegistered = validateRegisterStatus(replyStr);
+                natsMsg_Destroy(reply);
+            }
+            else
+            {
+                throw TelematicBridgeException(natsStatus_GetText(s));
```
paulbourelly999, line 41: Why do this while loop and throw and exception. If the registration fails you throw an exception so the loop is exited. If you are successful the loop is exited. Is there any condition in which you will retry. Only if NAT_OK and the validateRegisterStatus fails? In which situation would this occur.
\

paulbourelly999: Only if we get an invalid registration reply json from telematics?

src/v2i-hub/TelematicBridgePlugin/src/TelematicUnit.cpp
```diff
@@ -0,0 +1,314 @@
...
+    string TelematicUnit::constructAvailableTopicsReplyString(const unit_st &unit, const string &eventLocation, const string &testingType, const string &eventName, const vector<string> &availableTopicList, const string &excludedTopics)
+    {
+        Json::Value message;
+        message[UNIT_ID_KEY] = unit.unitId;
+        message[UNIT_NAME_KEY] = unit.unitName;
+        message[UNIT_TYPE_KEY] = unit.unitType;
+        message[LOCATION_KEY] = eventLocation;
+        message[TESTING_TYPE_KEY] = testingType;
+        message[EVENT_NAME_KEY] = eventName;
+        message[TIMESTAMP_KEY] = duration_cast<milliseconds>(system_clock::now().time_since_epoch()).count();
```
paulbourelly999, line 215: If you do not extend PluginClientClockAware and use the carma-clock object this plugin will not support simulation.

src/v2i-hub/TelematicBridgePlugin/src/TelematicUnit.cpp
dan-du-car: The server can return response but it does not have event information. It throw error when NATS_OK check fails. This usually because the server is down.

src/v2i-hub/TelematicBridgePlugin/src/TelematicUnit.cpp
dan-du-car: That is different case. We will address this later when we integrate this with simulation.

src/v2i-hub/TelematicBridgePlugin/REAADME.md
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/567/commits/71ac033b94eec7ea112cb5ead2d680a29b769772

src/v2i-hub/TelematicBridgePlugin/src/TelematicBridgePlugin.h
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/567/commits/71ac033b94eec7ea112cb5ead2d680a29b769772

src/v2i-hub/TelematicBridgePlugin/manifest.json
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/567/commits/71ac033b94eec7ea112cb5ead2d680a29b769772

src/v2i-hub/TelematicBridgePlugin/manifest.json
dan-du-car: This is a category defined by telematic tool. It only has two category platform and infrastructure. Carma-streets, carma-cloud, v2xhub belongs to infrastructure.

paulbourelly999: So then lets make this a constant and not configurable.

src/v2i-hub/TelematicBridgePlugin/manifest.json
dan-du-car: I think this unit id value can be anything that makes sense to end user during their testing with v2xhub as long as it is unique in the telematic tool.

src/v2i-hub/TelematicBridgePlugin/manifest.json
```diff
@@ -11,6 +11,31 @@
...
+            "description": "The NATS connection URL"            
+        },
+        {
+            "key": "UnitId",
+            "default": "v2xhub_id",
+            "description": "The unique identifier for a telematic unit"            
+        },
+        {
+            "key": "UnitName",
+            "default": "East Intersection",
```
paulbourelly999, line 27: Can this match J2735 intersection name or does it need to be separate. If it can match I suggest we just add in the description to set this to intersection name

dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/567/commits/1c59a3e2d0a95ea4ab44bc691b001aad589bd19c

src/v2i-hub/TelematicBridgePlugin/manifest.json
dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/567/commits/d5998cf26339e7da466a751de298804391f2ad1c

src/v2i-hub/TelematicBridgePlugin/manifest.json
paulbourelly999: So lets tell the user to use intersection name from MAP message if available.

src/v2i-hub/TelematicBridgePlugin/src/TelematicUnit.cpp
paulbourelly999: Is the response without event information also a bug or error behavior or would reattempts ever fix that. Also do we want to limit connection attempts on invalid responses?

paulbourelly999 (changes requested):
Left some more comments do be address

src/v2i-hub/TelematicBridgePlugin/src/TelematicUnit.cpp
dan-du-car: It is neither of them. It requires user attention to double check the event and its assigned unit.

dan-du-car: I think we want to wait for the valid response, otherwise there is no point to keep this plugin running.

src/v2i-hub/TelematicBridgePlugin/src/TelematicBridgePlugin.cpp
dan-du-car: This point is shared with this plugin. I cannot find a case where users will accidentally change the lifecycle of this pointer.

src/v2i-hub/TelematicBridgePlugin/manifest.json
paulbourelly999: Something like <rsu_21312>. We have not standardized this yet but I think if we say it has to be rsu_<map-intersection-id> then maybe we can minimize different ID

src/v2i-hub/TelematicBridgePlugin/manifest.json
paulbourelly999: Please specify J2735 MAP in the description. So should be set to MAP J2735 intersection name if available.

src/v2i-hub/TelematicBridgePlugin/src/TelematicBridgePlugin.cpp
paulbourelly999: If the plugin has a reference to this object but the telematics bridge  is no longer in scope, the telematics bridge will still not get deleted because the plugin holds a reference to this pointer which will prevent the pointer from getting deleted

paulbourelly999: Why does the plugin need a reference to this pointer?

src/v2i-hub/TelematicBridgePlugin/src/TelematicUnit.cpp
paulbourelly999: So if user input is required to make a valid response, then we should indicate to the user that something is wrong instead of silently indefinitely and continuously retrying a connection that has an issue that a user must resolve. 

To me either we do a single register attempt and fail or succeed and notify the user on fails, or we have a loop that attempts the connection some finite amount of times before failing. To me reattempts on connections should ideally be used for connection issues that do not require user input to be resolved, like if the server is restarting or if there are network temporary issues.

paulbourelly999: If the concern here is uptime of the resulting plugin, maybe this registration logic needs to be more complex and have some kind of failure recovery logic but this does not seem to me to satisfy that requirement

paulbourelly999 (changes requested):

src/v2i-hub/TelematicBridgePlugin/manifest.json
dan-du-car: The unit id is under the telematic tool context. They may share the same value. For debugging and ease of usability purpose, it might be more obvious if we keep both unit name and unit id in the same place like here on the V2xHub UI.

src/v2i-hub/TelematicBridgePlugin/src/TelematicBridgePlugin.cpp
dan-du-car: So that the plugin can use this pointer to send data into the telematic tool. It does not make sense for this plugin to subscribe messages but not forward them into the telematic tool.

src/v2i-hub/TelematicBridgePlugin/manifest.json
dan-du-car: This is not specific about MAP message or intersection. If we deploy this v2xhub in Cav-in-a-box at roadside, there is no concept of intersection. This east intersection is just a default value I give cause I test it with east intersection.

src/v2i-hub/TelematicBridgePlugin/src/TelematicUnit.cpp
dan-du-car: @paulbourelly999  Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/567/commits/040a99b1b4e251b04731c572cbc2da450bd0bcd0

src/v2i-hub/TelematicBridgePlugin/src/TelematicBridgePlugin.cpp
dan-du-car: @paulbourelly999  Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/567/commits/040a99b1b4e251b04731c572cbc2da450bd0bcd0

src/v2i-hub/TelematicBridgePlugin/manifest.json
paulbourelly999: I know this is not MAP Message specific from an implementation stand point but we should not be creating unnecessary new IDs in our different pieces of functionality. For example `INFRASTRUCTURE_ID` is a unique string used for registration for simulation but if we can reuse the ID for say telematics or if we could relate that ID to the unique J2735 Intersection ID we decrease the number of separately managed ID parameters. This way they all relate to each other.

dan-du-car: Can we table this for now? The unit id is something we used in the telematic tool context. The same design for this unit id is applied to carma-streets, carma-platform (vehicle), and carma-cloud. If this has some conflicts with IDs in a simulated environment, we probably want to decide and make the changes to all units to better fit the simulated environment.

dan-du-car: Do we have unique IDs for carma streets, carma-cloud , carma-platform as well to register for the simulation?

paulbourelly999 (approved):

PR #579, opened 2024-02-06
CDAR-756: Update UDPSockets to listen at higher frequency

paulbourelly999:
# PR Details
## Description
When running a simulation with CARMA Streets, the position data in the SDSMs appear to lag the ground truth significantly.
In the included image, the square inside the blue circle indicates the VRU’s reported position in the SDSM. The blue square is its actual position. They yellow circle shows CARMA’s reported SDSM position with the green rectangle showing its actual position.

image

Based on a visualization of carla-sensor-lib’s logs for the pedestrian (actor 221), it appears that the position data does get updated at some point. The image below is a timestamp vs. x position graph.
After further investigation, there is a difference in time stamps between the SDSMs and external objects.
image

It was found that is was largely due to the CDASimAdapter only consuming detection data at 10 hz (wall time). When running at real-time speed, the simulation can generate detection data at significantly higher frequency. This PR updates the consuming frequency to 200 Hz. This fix should be sufficient until the number of simultaneous detection exceeds 10 objects of 100Hz frequency.

Improved delay shown below:

screenshot

## Motivation and Context
Fix detection data delay for SDSM data

## How Has This Been Tested?
CDASim Deployment

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/sonar-scanner.yml
+++ b/.github/workflows/sonar-scanner.yml
@@ -20,6 +20,11 @@ jobs:
           # Disabling shallow clone is recommended for improving relevancy of reporting
           fetch-depth: 0
           submodules: recursive
+      - name: Set up JDK 11
+        uses: actions/setup-java@v3 # The setup-java action provides the functionality for GitHub Actions runners for Downloading and setting up a requested version of Java
+        with:
+          java-version: 17
+          distribution: "temurin"  
       - name:  Install sonar-scanner and build-wrapper
         uses: sonarsource/sonarcloud-github-c-cpp@v1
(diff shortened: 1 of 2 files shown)


MishkaMN (approved):
I have a little concern whether if we are introducing large frequency unnecessarily. 200hz seems high when the detections are coming in at 10hz, which we would need only 20Hz for. However, it is that high because all objects don't come at once, and if 5 objects are detected, they would need at least 100Hz (10Hz * 5 objects * 2 nyquist). Our current assumption is that the lidar or camera would provide objects in this way due to event driven action.

PR #593, opened 2024-04-03
CXC-28: Update CARMA Cloud Plugin to have configurable IP/Port

paulbourelly999:
# PR Details
## Description
Update CARMA Cloud plugin to allow for configurable target IP/Port for sending CARMA Cloud REST Requests
image
image

**NOTE**: Large changeset is due to correcting line ending characters for file. Feel free to double check change set in text diff tool of your choice. Here is what I got
image
image
image
image
image

## Related Issue
[CXC-28](https://usdot-carma.atlassian.net/browse/CXC-28)

## Motivation and Context
To support CARMA Cloud CDASim Integration testing

## How Has This Been Tested?
Locally integration tested using devcontainer to ensure configurable parameters were available and modified URL correctly

## Types of changes

## Checklist:

### DIFF
--- a/src/v2i-hub/CARMACloudPlugin/manifest.json
+++ b/src/v2i-hub/CARMACloudPlugin/manifest.json
@@ -29,6 +29,21 @@
 			"description":"Server IP address for V2X hub"
 		},
 		{
+			"key":"WebServicePort",
+			"default":"22222",
+			"description":"Server Port for V2X hub to receive TCM messages"
+		},
+		{
+			"key":"CARMACloudIP",
+			"default":"http://127.0.0.1",
+			"description":"Server IP address for V2X hub"
+		},
+		{
+			"key":"CARMACloudPort",
+			"default":"33333",
+			"description":"Server Port for V2X hub to receive TCM messages"
+		},
+		{
 			"key":"fetchTime",
 			"default":"15",
 			"description":"Time from while the TCMs are requested, in days"
@@ -62,11 +77,6 @@
 			"key": "TCMNOAcknowledgementDescription",
 			"default": "No response received from CMV after repeatedly broadcast TCMs.",
 			"description": "If the plugin does not receives any aknowledgement from CMV within the configured seconds that match the original TCM, the plugin will create an NO ACK message and display it on UI."
-		},
-		{
-			"key":"WebServicePort",
-			"default":"22222",
-			"description":"Server Port for V2X hub to receive TCM messages"
 		},
 		{
 			"key":"listTCM",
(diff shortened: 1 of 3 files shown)


src/v2i-hub/CARMACloudPlugin/manifest.json
```diff
@@ -28,6 +28,21 @@
 			"default":"127.0.0.1",
 			"description":"Server IP address for V2X hub"
 		},
+		{
+			"key":"WebServicePort",
```
kjrush, line 32: This name is a bit generic if this port is infact only for TCMs. Maybe rename it something more specific to reflect it's usage?

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
```diff
@@ -1,653 +1,657 @@
...
+	GetConfigValue<uint16_t>("TCMRepeatedlyBroadcastTimeOut",_TCMRepeatedlyBroadcastTimeOut);	
+	GetConfigValue<string>("TCMNOAcknowledgementDescription", _TCMNOAcknowledgementDescription);
+	GetConfigValue<int>("TCMRepeatedlyBroadCastTotalTimes", _TCMRepeatedlyBroadCastTotalTimes);
+	GetConfigValue<int>("TCMRepeatedlyBroadcastSleep", _TCMRepeatedlyBroadcastSleep);
+	GetConfigValue<string>("listTCM",list_tcm);
+	std::string carma_cloud_ip;
+	uint carma_cloud_port;
+	GetConfigValue<string>("CARMACloudIP",carma_cloud_ip);
+	GetConfigValue<uint>("CARMACloudPort",carma_cloud_port);
+	carma_cloud_url = carma_cloud_ip + ":" + std::to_string(carma_cloud_port);
```
kjrush, line 502: Is there a benefit to having these as separate parameters if we're going to concatenate them anyway? It seems like all it's doing is saving the user from having to know that ":" goes between the domain and the port. Would be simpler I think to just have a "carma_cloud_url" parameter in general?

dan-du-car (approved):

src/v2i-hub/CARMACloudPlugin/manifest.json
paulbourelly999: I can rename it but this is not the newly created configuration parameter for carma cloud port and ip. It is the existing one for the IP and port to deploy the CARMA-CLoud Plugin REST Server. Do you have any suggestions for better naming convention. V2X-Hub CARMA Cloud REST Server Port?

kjrush: Ah, if this is pre-existing I'm fine to leave it for now. But something like that, would be preferable. "Web Service Port" just leads me to ask "What web service"? and will likely get slightly more confusing when the new UI framework stuff is finalized, as there may be multiple web services running.

Issue #575, opened 2024-01-18
TelematicBridge: There is a memory leak in TelematicBridge plugin


dan-du-car:
### Summary

Run a load testing with telematic tool and v2xhub. The load testing is conducted in local by sending 10 different J2735 messages at 10hz to MessageReceiver plugin, and the testing lasts for around 10 hours. The TelematicBridge memory usage increases from around 0% to near 50% of total 15GB memory. 

### Version

4.3.0 (Current)

### Expected Behavior

The TelematicBridge memory usage should not increase indefinitely when it forwards messages to the telematic tool.

### Actual Behavior

The TelematicBridge memory usage increases.

### Steps to Reproduce the Actual Behavior

Run v2xhub
Run telematic tool in local 
Send sample J2735 messages to messagereceiver plugin at 10hz or higher frequency for a long time (more than 4 hours).
Update configuration to make sure telematicbridge is forwarding data generated by v2xhub to telematic tool.
Monitor the telematicbridge memory consumption

PR #599, opened 2024-04-09
TelematicPlugin: Memory leak

dan-du-car:
# PR Details
## Description
Fix memory leaks issues with several variables created with malloc or new.

## Related Issue
https://github.com/usdot-fhwa-OPS/V2X-Hub/issues/575

## How Has This Been Tested?
Local integration test

## Types of changes

## Checklist:

### DIFF
--- a/src/v2i-hub/TelematicBridgePlugin/src/TelematicBridgeMsgWorker.h
+++ b/src/v2i-hub/TelematicBridgePlugin/src/TelematicBridgeMsgWorker.h
@@ -109,6 +109,7 @@ namespace TelematicBridge
             // increase size of buffer.
             size_t new_size = 2 * (xb->allocated_size ? xb->allocated_size : 64);
             auto new_buf = static_cast<char *>(MALLOC(new_size));
+            // auto new_buf = new char[new_size];
             if (!new_buf)
                 return -1;
             // move old to new.
@@ -147,7 +148,9 @@ namespace TelematicBridge
         {
             throw TelematicBridgeException("Failed to  convert message with ID (=" + to_string(messageFrame->messageId) + ") to XML ");
         }
-        return string(xml_buffer.buffer);
+        auto output = string(xml_buffer.buffer);
+        FREEMEM(xml_buffer.buffer);
+        return output;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/v2i-hub/TelematicBridgePlugin/src/TelematicBridgeMsgWorker.h
```diff
@@ -238,7 +240,10 @@ namespace TelematicBridge
             }
             else
             {
-                json["payload"] = StringToJson(cJSON_Print(msg->payload));
+                // Render a cJSON entity to text for transfer/storage. Free the char* when finished.
+                char * payloadPtr = cJSON_Print(msg->payload);
```
paulbourelly999, line 244: Can we replace this with a smart pointer to remove the need for free?

paulbourelly999 (changes requested):
Please address sonar code smells and update PR description. It seems to be a copy of a previous issue.

src/v2i-hub/TelematicBridgePlugin/src/TelematicBridgeMsgWorker.h
dan-du-car: It only returns a char* Type. How do we put this in smart pointer?

paulbourelly999 (approved):

src/v2i-hub/TelematicBridgePlugin/src/TelematicBridgeMsgWorker.h
```diff
@@ -238,7 +240,9 @@ namespace TelematicBridge
             }
             else
             {
-                json["payload"] = StringToJson(cJSON_Print(msg->payload));
+                // Render a cJSON entity to text for transfer/storage. Free the char* when finished.
+                auto payloadPtr = std::make_unique<char*>(cJSON_Print(msg->payload));
```
paulbourelly999, line 244: Apparently `std::make_unique` is sort of a wrapper for `new`  (https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique) and actually creates a new object from the passed in parameter. This is likely why we still see the memory leak with this in. Apparently this is the more appropriate way to create a unique pointer from the returned char*
```cpp
 std::unique_ptr<char, decltype(std::free) *> t_copy { cJSON_Print(msg->payload), std::free };
```
Could you try this and see if it works?

paulbourelly999 (changes requested):
See review

paulbourelly999 (approved):
Looks good, what happened to the smart pointer stuff?

PR #602, opened 2024-04-11
CARMACloudPlugin: Fix the issue to unzip the compressed TCMs

dan-du-car:
# PR Details
## Description
Fix issue with CARMACloudPlugin unzip the compressed TCM payload. The unzip function did not append the correct numbers of characters. Add 32 to WinBits to allow header detection. Remove toLocal8Bit() conversion as the xml is by default sent as utf-8 encoding.

## Related Issue
NA

## How Has This Been Tested?
Local integration test

## Types of changes

## Checklist:

### DIFF
--- a/src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
+++ b/src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
@@ -240,6 +240,7 @@ void CARMACloudPlugin::CARMAResponseHandler(QHttpEngine::Socket *socket)
 	tcm=updateTags(tcm,"TrafficControlParams","params");
 	tcm=updateTags(tcm,"TrafficControlGeometry","geometry");
 	tcm=updateTags(tcm,"TrafficControlPackage","package");
+	PLOG(logDEBUG2) << "Received TCM: " << tcm << std::endl;
 
 	std::list<std::string> tcmSL = {};
 	if (isCompressed)
--- a/src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
+++ b/src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
@@ -208,6 +208,7 @@ void CARMACloudPlugin::CARMAResponseHandler(QHttpEngine::Socket *socket)
 	QByteArray st; 
 	while(socket->bytesAvailable()>0)
 	{	
+		PLOG(logDEBUG) << "Bytes available." << std::endl;
 		auto readBytes = socket->readAll();
 		st.append(readBytes);
 	}
@@ -223,7 +224,12 @@ void CARMACloudPlugin::CARMAResponseHandler(QHttpEngine::Socket *socket)
 	bool isCompressed = socket->headers().keys().contains(CONTENT_ENCODING_KEY) && std::string(socket->headers().constFind(CONTENT_ENCODING_KEY).value().data()) == CONTENT_ENCODING_VALUE;
 	if (isCompressed)
 	{
-		tcm = UncompressBytes(st).data();
+		QByteArray tcmBytes = UncompressBytes(st);
+		if(tcmBytes.size() == 0)
+		{
+			return;
+		}
+		tcm = tcmBytes.data();
 	}else{
 		tcm = st.data();		
 	}
@@ -235,15 +241,15 @@ void CARMACloudPlugin::CARMAResponseHandler(QHttpEngine::Socket *socket)
 	tcm=updateTags(tcm,"TrafficControlGeometry","geometry");
 	tcm=updateTags(tcm,"TrafficControlPackage","package");
 
-	std::list<std::string> tcm_sl = {};
+	std::list<std::string> tcmSL = {};
 	if (isCompressed)
 	{
-		tcm_sl = FilterTCMs(tcm);
+		tcmSL = FilterTCMs(tcm);
 	}else{
-		tcm_sl.push_back(tcm);
+		tcmSL.push_back(tcm);
 	}
 
-	for(const auto tcm_s: tcm_sl)
+	for(const auto tcm_s: tcmSL)
 	{
 		tsm5Message tsm5message;
 		tsm5EncodedMessage tsm5ENC;
@@ -628,7 +634,7 @@ QByteArray CARMACloudPlugin::UncompressBytes(const QByteArray compressedBytes) c
             outBuf.append(buffer, BUFFER_SIZE - strm.avail_out);
 		} while (Z_STREAM_END != isDone); // Reach the end of stream to be uncompressed
 	}else{
-		PLOG(logWARNING) << "Error initalize stream. Err code = " << err << std::endl;
+		PLOG(logERROR) << "Error initalize stream. Err code = " << err << std::endl;
 	}
 	//Finished decompress data stream
     inflateEnd(&strm);
--- a/src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
+++ b/src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
@@ -219,7 +219,7 @@ void CARMACloudPlugin::CARMAResponseHandler(QHttpEngine::Socket *socket)
 	}
 	PLOG(logINFO) << "Received TCM bytes size: " << st.size()<< std::endl;
 
-	string tcm = "";
+	std::string tcm = "";
 	bool isCompressed = socket->headers().keys().contains(CONTENT_ENCODING_KEY) && std::string(socket->headers().constFind(CONTENT_ENCODING_KEY).value().data()) == CONTENT_ENCODING_VALUE;
 	if (isCompressed)
 	{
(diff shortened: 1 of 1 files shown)


src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
```diff
@@ -205,49 +205,50 @@ string CARMACloudPlugin::updateTags(string str,string tagout, string tagin)
...
 	tcm=updateTags(tcm,"TrafficControlGeometry","geometry");
 	tcm=updateTags(tcm,"TrafficControlPackage","package");
-	
-	//List of tcm in string format
-	std::list<std::string> tcm_sl = FilterTCMs(tcm);	
+
+	std::list<std::string> tcm_sl = {};
+	if (isCompressed)
+	{
+		tcm_sl = FilterTCMs(tcm);
```
kjrush, line 241: Is there a reason we only call FilterTCMs if the message was originally compressed? If so, it's probably worth at least a comment.

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
```diff
@@ -604,13 +605,13 @@ void CARMACloudPlugin::ConvertString2Pair(std::pair<string,string> &str_pair, co
 QByteArray CARMACloudPlugin::UncompressBytes(const QByteArray compressedBytes) const
 {
     z_stream strm;
-	strm.zalloc = nullptr;//Refer to zlib docs (https://zlib.net/zlib_how.html)
-	strm.zfree = nullptr; 
-    strm.opaque = nullptr;
+	strm.zalloc = Z_NULL;//Refer to zlib docs (https://zlib.net/zlib_how.html)
```
kjrush, line 608: Whitespace in this function looks a little weird, but could just be github's rendering.

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
```diff
@@ -205,49 +205,50 @@ string CARMACloudPlugin::updateTags(string str,string tagout, string tagin)
...
-	if(tcm.length() == 0)
+	if(st.size() == 0)
 	{
-		PLOG(logERROR) << "Received TCM length is zero, and skipped." << std::endl;
+		PLOG(logERROR) << "Received TCM is empty, and skipped." << std::endl;
 		return;
 	}
+	PLOG(logINFO) << "Received TCM bytes size: " << st.size()<< std::endl;
+
+	string tcm = "";
```
kjrush, line 222: Just for consistency's sake, is there a reason we use `string` here and `std::string` below?

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
dan-du-car: The compressed file has list of TCMs wrapped inside the <TrafficControlMessageList> tags. All these are in a big XML payload. This filter tcms is to remove the list tags and only return a list a TCMs objects. There are comments in the header file: https://github.com/usdot-fhwa-OPS/V2X-Hub/blob/625800ebaa77466a5e2eb7a584b57e7f9bce263e/src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.h#L169

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
dan-du-car: updated

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
paulbourelly999: So if not compressed it is guranteed to be a list of a single TCM but if it is compressed it is multiple TCMs in a single XML that have to be parsed out individually? Is that correct?

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
```diff
@@ -623,11 +624,10 @@ QByteArray CARMACloudPlugin::UncompressBytes(const QByteArray compressedBytes) c
             char buffer[BUFFER_SIZE] = {0};
             strm.avail_out = BUFFER_SIZE;
             strm.next_out = (Byte *)buffer;
-			//Uncompress finished
-            isDone = inflate(&strm, Z_FINISH);
-            outBuf.append(buffer);
-        } while (Z_STREAM_END != isDone); //Reach the end of stream to be uncompressed 
-    }else{
+            isDone = inflate(&strm, Z_NO_FLUSH);
+			outBuf.append(buffer, BUFFER_SIZE - strm.avail_out);
```
paulbourelly999, line 628: Nit: Indentation looks off.

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
```diff
@@ -376,7 +377,7 @@ void CARMACloudPlugin::TCMAckCheckAndRebroadcastTCM()
 		}
 		else
 		{
-			PLOG(logDEBUG) << "NO TCMs to broadcast." << std::endl;
+			PLOG(logDEBUG4) << "NO TCMs to broadcast." << std::endl;
```
paulbourelly999, line 380: Would it be useful to have a status that evaluates to the number of TCMs currently broadcasting rather than this log statement.

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
paulbourelly999: So then what is this configuration parameter for https://github.com/usdot-fhwa-OPS/V2X-Hub/blob/0e1ee6502a6591769a59e4c08d88262d89aa3d52/src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp#L94

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
```diff
@@ -623,11 +624,10 @@ QByteArray CARMACloudPlugin::UncompressBytes(const QByteArray compressedBytes) c
...
-			//Uncompress finished
-            isDone = inflate(&strm, Z_FINISH);
-            outBuf.append(buffer);
-        } while (Z_STREAM_END != isDone); //Reach the end of stream to be uncompressed 
-    }else{
+            isDone = inflate(&strm, Z_NO_FLUSH);
+            outBuf.append(buffer, BUFFER_SIZE - strm.avail_out);
+		} while (Z_STREAM_END != isDone); // Reach the end of stream to be uncompressed
+	}else{
 		PLOG(logWARNING) << "Error initalize stream. Err code = " << err << std::endl;
```
paulbourelly999, line 631: Do we still want to execute the last line of this method if this error happens?

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
```diff
@@ -623,11 +624,10 @@ QByteArray CARMACloudPlugin::UncompressBytes(const QByteArray compressedBytes) c
...
             strm.avail_out = BUFFER_SIZE;
             strm.next_out = (Byte *)buffer;
-			//Uncompress finished
-            isDone = inflate(&strm, Z_FINISH);
-            outBuf.append(buffer);
-        } while (Z_STREAM_END != isDone); //Reach the end of stream to be uncompressed 
-    }else{
+            isDone = inflate(&strm, Z_NO_FLUSH);
+            outBuf.append(buffer, BUFFER_SIZE - strm.avail_out);
+		} while (Z_STREAM_END != isDone); // Reach the end of stream to be uncompressed
```
paulbourelly999, line 629: Is there a reason for do while instead of while loop. Just curious

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
dan-du-car: Yeah, that might be a nice feature to add. This is an indicator that no more TCMs that is  left to broadcast.

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
dan-du-car: It seems useful to know that the stream is not readable and what the actual error is. What do you suggest?

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
dan-du-car: Because it needs to read the data at least once to know if isDone reading all data or not?

src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
paulbourelly999: I am asking whether we want the statements on L634 and L635 to execute in this error case. I have no problem with the logging except if it is an error we may want to make it error level and not warning? Just curious whether we want to stop doing everything in this case or not.

dan-du-car: Updated it to error type and handled it silently by ignoring it for now.

paulbourelly999 (approved):

Issue #603, opened 2024-04-12
Intermittent TCM Reception Failure in V2XHub CARMACloudPlugin


chengyuan0124:
### Summary

We have observed inconsistent behavior with the TCM payload reception in the CARMACloudPlugin of V2XHub. While the plugin sometimes successfully receives TCM data, there are instances where it fails to report a TCM received. This issue has been confirmed through multiple tests, including recent updates made to the code and associated images. Despite these efforts, the problem persists, and the root cause remains unidentified. We need to investigate the HTTP server’s behavior in the plugin to ensure reliable data handling

### Commit

NA

### Expected Behavior

The HTTP server in the CARMACloudPlugin should consistently receive and process TCM payloads sent to V2XHub. Whenever TCM data is transmitted to the plugin, it should be properly received and reported without interruptions or failures, ensuring reliable data handling and system functionality.

### Actual Behavior

Currently, the HTTP server in the CARMACloudPlugin exhibits intermittent behavior in receiving TCM payloads. While it sometimes succeeds in receiving and processing the data, there are frequent occurrences where it fails to report a TCM received. 

### Steps to Reproduce the Actual Behavior

NA

Issue #601, opened 2024-04-11
Intermittent TCM Reception and Processing Issue in V2XHub with Combined XML Blocks


chengyuan0124:
### Summary

We are facing a multifaceted issue with V2XHub, where it intermittently fails to receive TCMs sent from CARMA-Cloud and also does not process combined TCM XML blocks correctly. Initial suspicions pointed towards network setting issues

### Commit

NA

### Expected Behavior

V2X-Hub can receive and unzip TCM sent from CARMA-Cloud

### Actual Behavior

V2X-Hub fail to receive and unzip TCM sent from CARMA-Cloud

### Steps to Reproduce the Actual Behavior

NA

chengyuan0124:
Issue has been addressed by @dan-du-car

PR #604, opened 2024-04-15
Fix tcm intermittent reception

dan-du-car:
# PR Details
## Description
Fix the TCM intermittent reception issue where carma cloud sends TCMs in an http request and v2xhub CARMACloudPlugin sometimes does not receive the TCMs. Replace the OpenAI generated code with QhttpEngine recommended functional call, refer to: https://ci.quickmediasolutions.com/job/qhttpengine-documentation/doxygen/classQHttpEngine_1_1QObjectHandler.html.

## Related Issue
https://github.com/usdot-fhwa-OPS/V2X-Hub/issues/601
https://github.com/usdot-fhwa-OPS/V2X-Hub/issues/603

## Motivation and Context
CARMA cloud integration testing

## How Has This Been Tested?
Local integration test

## Types of changes

## Checklist:

### DIFF
--- a/src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
+++ b/src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp
@@ -463,25 +463,14 @@ int CARMACloudPlugin::StartWebService()
  	QHostAddress address = QHostAddress(QString::fromStdString (webip));
 	quint16 port = static_cast<quint16>(webport);
 
-	QSharedPointer<OpenAPI::OAIApiRequestHandler> handler(new OpenAPI::OAIApiRequestHandler());
-	handler = QSharedPointer<OpenAPI::OAIApiRequestHandler> (new OpenAPI::OAIApiRequestHandler());
+	QHttpEngine::QObjectHandler apiHandler;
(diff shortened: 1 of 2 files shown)


paulbourelly999 (changes requested):
Address code smells. It is also not great practice to manually edit generated code but if it is necessary I guess it is not the only place we do this. Lastly do we know at all why the previous implementation did not work? Without knowing that how have we confirmed that this implementation fixes the issue?

@paulbourelly999 Bryan and I both tested this change. It worked. It is not an edit of the generated code, rather a complete replacement of it with the use the QhttpEngine implementation. I think the generated code did not work for some time now because I receive reports on the message loss from multiple plugins that use this generated code.

jwillmartin (approved):
Using official documentation/guidance is more appropriate than generated code. This looks good and I see your testing passed.

jwillmartin (commented):
My bad, I see there were some code smells when I opened the link to sonar cloud. I agree with Paul on addressing those.

paulbourelly999:
@dan-du-car Wait so the edits are not on the generated code, That makes sense. You mentioned it is a replacement of generated code. Does that mean some of the generated code can be removed? You also mentioned that this is an issue with other OpenAPI generated REST clients/servers. If so we should create issues to remove the problematic code and include any evidence you have of these other services having similar issues.

Issue #608, opened 2024-04-18
MAP Plugin crashes with segmentation fault


jwillmartin:
### Summary

The MAP plugin crashes when a MAP message is opened to be read. This only happens when a MAP message contains a newline character. 

Most text editors automatically add a newline character at the end of files for human readability. But this causes an issue in the case of the plugin reading all contents. 

### Version

4.3.0 (Current)

### Expected Behavior

The MAP plugin is able to read a file and decode, then encode the contents into a MAP message. 

### Actual Behavior

MAP plugin reads file and crashes is file contains a newline.

### Steps to Reproduce the Actual Behavior

1. Disable MAP Plugin
2. Change configuration to open a file with MapData contents
3. Add a newline character to the end of the file
4. Enable MAP plugin

PR #609, opened 2024-04-18
Fix map

jwillmartin:
# PR Details
## Description

Changes were made to handle files that could be expected. This includes:
Files with newline characters
Files with only MapData
Files containing a MessageFrame along with MapData

Plugin was also cleaned up with a header file. 

## Related Issue

#608

## Motivation and Context
These changes make the MAP plugin more robust because it's able to handle MAP files that could be expected to be generated by a user. 

## How Has This Been Tested?

Tested using files with and without newline characters, files with and without MessageFrames, and a combination of the two. 

## Types of changes

- [X] Defect fix (non-breaking change that fixes an issue)

## Checklist:

- [X] All new and existing tests passed.


src/v2i-hub/MapPlugin/src/MapPlugin.cpp
```diff
@@ -342,43 +245,64 @@ bool MapPlugin::LoadMapFiles()
...
+							PLOG(logERROR) << "Failed to open file: " << fn;
+						} else {
+							try {
+								std::string fileContent((std::istreambuf_iterator<char>(in)), std::istreambuf_iterator<char>());
+								in.close();
+								// Remove any newline characters
+								fileContent.erase(remove(fileContent.begin(), fileContent.end(), '\n'), fileContent.end());
+
+								// Check for and remove MessageFrame
+								if (fileContent.size() >= 4 && fileContent.substr(0, 4) == "0012") {
```
dan-du-car, line 259: What happens if there are multiple "38" in the binary payload? The same for 0012?

src/v2i-hub/MapPlugin/src/MapPlugin.cpp
```diff
@@ -342,43 +245,64 @@ bool MapPlugin::LoadMapFiles()
...
+									size_t pos = fileContent.find("38");
+									PLOG(logDEBUG) << "Beginning of MapData found at: " << pos;
+									fileContent.erase(0, pos);
```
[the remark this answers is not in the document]
dan-du-car: I would recommend to move this file reading and sanitizing steps into a separate function.

src/v2i-hub/MapPlugin/src/MapPlugin.h
```diff
@@ -0,0 +1,104 @@
...
+	std_attribute(this->msg, int, Action, -1, );
+	std_attribute(this->msg, std::string, FilePath, "", );
+	std_attribute(this->msg, std::string, InputType, "", );
+	std_attribute(this->msg, std::string, Bytes, "", );
+
+	static tmx::message_tree_type to_tree(MapFile m) {
+		return tmx::message::to_tree(static_cast<tmx::message>(m));
+	}
+
+	static MapFile from_tree(tmx::message_tree_type tree) {
```
dan-du-car, line 64: Are these from_tree and to_tree functions used?

src/v2i-hub/MapPlugin/src/MapPlugin.h
```diff
@@ -0,0 +1,104 @@
...
+#endif
+
+class MapFile: public tmx::message {
+public:
+	MapFile(): tmx::message() {}
+	virtual ~MapFile() {}
+
+	std_attribute(this->msg, int, Action, -1, );
+	std_attribute(this->msg, std::string, FilePath, "", );
+	std_attribute(this->msg, std::string, InputType, "", );
```
dan-du-car, line 57: Is this "inputType" used? What does this represent?

src/v2i-hub/MapPlugin/src/MapPlugin.h
```diff
@@ -0,0 +1,104 @@
...
+#if SAEJ2735_SPEC < 63
+UPERframe _uperFrameMessage;
+#endif
+
+class MapFile: public tmx::message {
+public:
+	MapFile(): tmx::message() {}
+	virtual ~MapFile() {}
+
+	std_attribute(this->msg, int, Action, -1, );
```
dan-du-car, line 55: What is the rest of actions this MAP support? Which default action is this initialized to?

src/v2i-hub/MapPlugin/src/MapPlugin.cpp
dan-du-car: Great findings on these bugs!

jwillmartin: @dan-du-car The message I used had 6 "38" in it. It only looked for the first instance, so this worked. But I can change it to make sure it only looks for it in the first couple of bytes depending on payload length.

src/v2i-hub/MapPlugin/src/MapPlugin.cpp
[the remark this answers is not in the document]
jwillmartin: Good idea, I'll move it

src/v2i-hub/MapPlugin/src/MapPlugin.h
jwillmartin: Doesn't look like it. They were also included in another the inputs/ISDDataAdaptor.hpp and also not used. I'll remove it and retest

src/v2i-hub/MapPlugin/src/MapPlugin.h
jwillmartin: Looks like it was declared but is expected to be empty? I'll find out if we need it or not

src/v2i-hub/MapPlugin/src/MapPlugin.cpp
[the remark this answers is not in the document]
jwillmartin: It removes the number of pos bytes, starting from 0. So in this case removes bytes 0:[pos-1]

src/v2i-hub/MapPlugin/src/MapPlugin.cpp
jwillmartin: I'll leave it as is. Find only looks for the first instance of the string.

src/v2i-hub/MapPlugin/src/MapPlugin.h
jwillmartin: It was needed by message.hpp in TmxApi. Left it in

src/v2i-hub/MapPlugin/src/MapPlugin.h
jwillmartin: It's needed to find the file extension type for ISD, TXT, XML, JSON, UPER. Kept it in

src/v2i-hub/MapPlugin/src/MapPlugin.h
jwillmartin: Default is no action existing. It looks like only action is for finding the file path.

src/v2i-hub/MapPlugin/src/MapPlugin.h
dan-du-car: Cool. The difference between UPER and TXT upload? It seems they both upload bytes but TXT has message frame in it? Why do we need these two different uploads that has similar content? I am not familiar with the usage of this plugin, so I am a bit curious about these different extension types?

src/v2i-hub/MapPlugin/src/MapPlugin.cpp
```diff
@@ -304,6 +207,17 @@ int MapPlugin::Main() {
 	return (EXIT_SUCCESS);
 }
 
+string MapPlugin::removeMessageFrame(string &fileContent) {
+    // Check for and remove MessageFrame
+	if (fileContent.size() >= 4 && fileContent.substr(0, 4) == "0012") {
```
paulbourelly999, line 212: Pull from MAP PSID constant

dan-du-car (approved):

paulbourelly999 (changes requested):
I requested some changes here. Please let me know if this story/PR is growing more in scope and needs some kind of story associated with it. I can help implementing the changes as well if needed.

src/v2i-hub/MapPlugin/src/MapPlugin.cpp
```diff
@@ -304,6 +207,39 @@ int MapPlugin::Main() {
...
+	string map_messageID = mapID_buffer;
+	return map_messageID;
+}
+
+string MapPlugin::removeMessageFrame(string &fileContent)
+{
+    string map_messageID = enum_to_hex_string();
+
+    // Check for and remove MessageFrame
+    if (fileContent.size() >= 4 && fileContent.substr(0, 4) == map_messageID)
```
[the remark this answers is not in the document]
jwillmartin: We could do that, but we'd still have to have logic to figure out whether the header has an additional 2 bytes or not. The message frame will always contain the message ID, at the beginning so this is a good way to quickly check for it instead of trying and failing to encode the message. I think both ways would work, but I'd rather run this check instead of attempting a blind encoding.

src/v2i-hub/MapPlugin/src/MapPlugin.h
```diff
@@ -0,0 +1,111 @@
...
+	J2735MessageFactory factory;
+
+	int sendFrequency = 1000;
+	FrequencyThrottle<int> errThrottle;
+
+	char mapID_buffer[5];
+
+	bool LoadMapFiles();
+	void DebugPrintMapFiles();
+	string enum_to_hex_string();
```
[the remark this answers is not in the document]
jwillmartin: It's to convert the messageID from a decimal enum to a hex string. The only place the message ID is declared in [TmxApiMessages](https://github.com/usdot-fhwa-OPS/V2X-Hub/blob/develop/src/tmx/TmxApi/tmx/TmxApiMessages.h#L76)

src/v2i-hub/MapPlugin/src/MapPlugin.h
```diff
@@ -0,0 +1,111 @@
...
+	string removeMessageFrame(string &fileContent);
```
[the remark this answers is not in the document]
jwillmartin: Responded in other comment

src/v2i-hub/MapPlugin/src/MapPlugin.h
```diff
@@ -0,0 +1,111 @@
+#pragma once
```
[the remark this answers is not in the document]
jwillmartin: Kept is as #pragma once, thanks!

src/v2i-hub/MapPlugin/src/MapPlugin.cpp
```diff
@@ -1,114 +1,17 @@
...
-#include "utils/common.h"
-#include "utils/map.h"
-
-#include <MapSupport.h>
-using namespace std;
-using namespace tmx;
-using namespace tmx::messages;
-using namespace tmx::utils;
+#include "MapPlugin.h"
+#include <chrono>
```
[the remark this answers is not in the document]
jwillmartin: Good catch, moved over to header.

src/v2i-hub/MapPlugin/src/MapPlugin.cpp
```diff
@@ -304,6 +207,39 @@ int MapPlugin::Main() {
 	return (EXIT_SUCCESS);
 }
 
+string MapPlugin::enum_to_hex_string()
+{
+	sprintf(mapID_buffer, "%04X", tmx::messages::api::mapData);
```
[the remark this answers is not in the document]
jwillmartin: Initialized with zeros in header file. Missed the {0}, but just added it.

src/v2i-hub/MapPlugin/src/MapPlugin.cpp
```diff
@@ -304,6 +207,39 @@ int MapPlugin::Main() {
...
+	string map_messageID = mapID_buffer;
+	return map_messageID;
+}
+
+string MapPlugin::removeMessageFrame(string &fileContent)
```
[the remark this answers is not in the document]
jwillmartin: fileContent is passed by reference and the method is only called after fileContent has already been initialized. Let me know if you want to add anything else to it to make sure this method doesn't crash the plugin

src/v2i-hub/MapPlugin/src/MapPlugin.cpp
```diff
@@ -304,6 +204,39 @@ int MapPlugin::Main() {
...
```
paulbourelly999, line 214: This should be `const string &filcontent`

src/v2i-hub/MapPlugin/src/MapPlugin.h
paulbourelly999: I think any format we accept should very message content before broadcast. When it is not verified we get exceptions on down stream plugins that attempt to ingest this message. I created the TXT format a while back to avoid dealing with issues with the UPER format logic. These two formats can likely be consolidated since they are both uper hex

paulbourelly999 (changes requested):
There are a couple more things I would like fixed inside the MAP plugin under this issue as long as this issue is not blocking some other work or stakeholder. I think we should consolidate TXT and UPER formats and avoid any direct byte reading or changing. We should verify any message and possibly log (tmx-event) if we are unable to verify a given message.
If you believe we can do some of this later, if you could create a follow on story for the MAP plugin to consolidate these two formats that would be create. Lastly I just want to confirm that both TXT and UPER have been tested for functionality before merging this.

paulbourelly999 (approved):

PR #613, opened 2024-05-30
Add configuration parameters for RSUHealthMonitorPlugin

dan-du-car:
# PR Details
## Description
Add configuration parameter to set SNMP security level.

## Related Issue
https://github.com/usdot-fhwa-stol/cda-telematics/issues/213

## JIRA
https://usdot-carma.atlassian.net/browse/VH-1316
## Motivation and Context
Telematic tool integration test

## How Has This Been Tested?
Unit test

## Types of changes

## Checklist:

### DIFF
--- a/src/v2i-hub/RSUHealthMonitorPlugin/manifest.json
+++ b/src/v2i-hub/RSUHealthMonitorPlugin/manifest.json
@@ -19,7 +19,7 @@
         },
         {
             "key":"RSUConfigurationList",
-            "default":"{ \"RSUS\": [ { \"RSUIp\": \"192.168.XX.XX\", \"SNMPPort\": \"161\", \"AuthPassPhrase\": \"dummy\", \"User\": \"authOnlyUser\", \"RSUMIBVersion\": \"RSU4.1\" },{ \"RSUIp\": \"192.168.00.XX\", \"SNMPPort\": \"162\", \"AuthPassPhrase\": \"tester\", \"User\": \"authPrivUser\", \"RSUMIBVersion\": \"RSU4.1\" }] }",
+            "default":"{ \"RSUS\": [ { \"RSUIp\": \"192.168.XX.XX\", \"SecurityLevel\":\"authPriv\", \"SNMPPort\": \"161\", \"AuthPassPhrase\": \"dummy\", \"User\": \"authOnlyUser\", \"RSUMIBVersion\": \"RSU4.1\" ,\"RSUIdentifier\":\"NA\"},{ \"RSUIp\": \"192.168.00.XX\", \"SecurityLevel\":\"authPriv\", \"SNMPPort\": \"162\", \"AuthPassPhrase\": \"tester\", \"User\": \"authPrivUser\", \"RSUMIBVersion\": \"RSU4.1\" ,\"RSUIdentifier\":\"NA\"}] }",
             "description":"Configurations of the RSUs the V2X hub is connected to."
         }
     ]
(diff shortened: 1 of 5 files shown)


src/v2i-hub/RSUHealthMonitorPlugin/manifest.json
```diff
@@ -19,7 +19,7 @@
         },
         {
             "key":"RSUConfigurationList",
-            "default":"{ \"RSUS\": [ { \"RSUIp\": \"192.168.XX.XX\", \"SNMPPort\": \"161\", \"AuthPassPhrase\": \"dummy\", \"User\": \"authOnlyUser\", \"RSUMIBVersion\": \"RSU4.1\" },{ \"RSUIp\": \"192.168.00.XX\", \"SNMPPort\": \"162\", \"AuthPassPhrase\": \"tester\", \"User\": \"authPrivUser\", \"RSUMIBVersion\": \"RSU4.1\" }] }",
+            "default":"{ \"RSUS\": [ { \"RSUIp\": \"192.168.XX.XX\", \"SecurityLevel\":\"authPriv\", \"SNMPPort\": \"161\", \"AuthPassPhrase\": \"dummy\", \"User\": \"authOnlyUser\", \"RSUMIBVersion\": \"RSU4.1\" ,\"RSUIdentifier\":\"NA\"},{ \"RSUIp\": \"192.168.00.XX\", \"SecurityLevel\":\"authPriv\", \"SNMPPort\": \"162\", \"AuthPassPhrase\": \"tester\", \"User\": \"authPrivUser\", \"RSUMIBVersion\": \"RSU4.1\" ,\"RSUIdentifier\":\"NA\"}] }",
```
paulbourelly999, line 22: Is RSU Name or ID not the identifier? I think the RSU spec has snmp endpoints for both.

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorPlugin.cpp
```diff
@@ -27,6 +27,7 @@ namespace RSUHealthMonitor
         for (auto rsuConfig : _rsuConfigListPtr->getConfigs())
         {
             auto rsuStatusJson = _rsuWorker->getRSUStatus(rsuConfig.mibVersion, rsuConfig.rsuIp, rsuConfig.snmpPort, rsuConfig.user, rsuConfig.authPassPhrase, rsuConfig.securityLevel, SEC_TO_MICRO);
+            rsuStatusJson[RSUIdentifierKey] = rsuConfig.RSUIdentifier;
```
paulbourelly999, line 30: Why are we pulling this from the configuration, are we not setting this on the RSU?

paulbourelly999 (changes requested):
One issue I have with the current setup is that is seems we are creating our own RSU Identifier instead of using the existing RSU name and/or ID fields both of which can be set and retrieved from the RSU via SNMP. 

@paulbourelly999  Removed

src/v2i-hub/RSUHealthMonitorPlugin/manifest.json
dan-du-car: RSUs specs does have RSU Name and ID. However, they usually have the same values across RSUs. We could use the RSUs specs, and that requires to update each individual RSUs setup either via SNMP request or some other ways. For the telematic tool purpose, I assume this is the least effort to meet the telematic requirement.

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUHealthMonitorPlugin.cpp
dan-du-car: This is to directly sent as part of the RSU status payload.

paulbourelly999 (approved):

PR #616, opened 2024-06-10
Optional SENSOR_CONFIG_JSON_FILE

paulbourelly999:
# PR Details
## Description
This 

## Types of changes

## Checklist:

### DIFF
--- a/src/tmx/TmxUtils/src/simulation/SimulationEnvUtils.cpp
+++ b/src/tmx/TmxUtils/src/simulation/SimulationEnvUtils.cpp
@@ -13,7 +13,7 @@ namespace tmx::utils::sim{
 
     }
 
-    std::string get_sim_config(const char *config_name) {
+    std::string get_sim_config(const char *config_name, bool required = true) {
         if (is_simulation_mode() && config_name) {
             try {
                 std::string config =  std::getenv(config_name);
@@ -21,7 +21,9 @@ namespace tmx::utils::sim{
             }
             catch(const std::logic_error &e) {
                 std::string config_name_str = config_name;
-                throw TmxException("Simulation Config " + config_name_str + " not set!");
+                if ( required ) {
+                    throw TmxException("Required simulation config " + config_name_str + " not set!");
+                }
             }
         } else {
             throw TmxException("V2X-Hub not in sumulation mode or config param name is null pointer!");
--- a/src/v2i-hub/CDASimAdapter/src/CDASimConnection.cpp
+++ b/src/v2i-hub/CDASimAdapter/src/CDASimConnection.cpp
@@ -61,6 +61,9 @@ namespace CDASimAdapter{
             }     
             message["sensors"] = sensors_json_v;
         }
+        else {
+            PLOG(logWARNING) << "No sensors where configured for this V2X-Hub instance."
+        }
         Json::StyledWriter writer;
         message_str = writer.write(message);
         return message_str;
--- a/src/v2i-hub/CDASimAdapter/src/CDASimConnection.cpp
+++ b/src/v2i-hub/CDASimAdapter/src/CDASimConnection.cpp
@@ -6,7 +6,7 @@ using namespace tmx::utils;
 namespace CDASimAdapter{ 
     CDASimConnection::CDASimConnection(const std::string &simulation_ip, const std::string &infrastructure_id, const uint simulation_registration_port, const uint sim_v2x_port,
                                                         const std::string &local_ip,  const uint time_sync_port,const uint simulated_interaction_port, const uint v2x_port, 
-                                                        const Point &location, const std::string &sensor_json_file_path) : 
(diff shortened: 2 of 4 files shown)


src/v2i-hub/CDASimAdapter/src/CDASimConnection.cpp
```diff
@@ -52,12 +52,18 @@ namespace CDASimAdapter{
...
-        }     
-        message["sensors"] = sensors_json_v;
+        // Sensor configuration is an optional part of registration message.
+        if ( !_sensor_json_file_path.empty() ) {
+            auto sensors_json_v = read_json_file(_sensor_json_file_path);
+            if(sensors_json_v.empty())
+            {
+                PLOG(logWARNING) << "Sensors JSON is empty!" << std::endl;
+            }     
+            message["sensors"] = sensors_json_v;
```
dan-du-car, line 62: Do we need to set this sensors in message? I noticed that if the path to sensor json file is empty, this sensors is not set in message.

src/tmx/TmxUtils/src/simulation/SimulationEnvUtils.cpp
```diff
@@ -13,15 +13,17 @@ namespace tmx::utils::sim{
 
     }
 
-    std::string get_sim_config(const char *config_name) {
+    std::string get_sim_config(const char *config_name, bool required) {
         if (is_simulation_mode() && config_name) {
             try {
                 std::string config =  std::getenv(config_name);
```
dan-du-car, line 19: Is this the getenv() used? https://en.cppreference.com/w/cpp/utility/program/getenv#:~:text=This%20function%20is%20not%20required,obtained%20from%20a%20previous%20call. It seems to return null pointer when the config name not found?

paulbourelly999: See unit tests. It throws a std::logic_error when configuration parameter does not exist.

src/v2i-hub/CDASimAdapter/src/CDASimConnection.cpp
paulbourelly999: Yep it is an optional part of the registration message. When included it configures simulated sensor for V2X-Hub/CARMA-Streets. Since V2X-Hub/CARMA-Streets basic functionality does not require sensors it is also not required.

src/tmx/TmxUtils/src/simulation/SimulationEnvUtils.cpp
paulbourelly999: Actually I think you are right. Something about trying to assign a nullptr to a string or return it throws std::logic_error. I will fix this logic to explicitly confirm it is no null ptr.

paulbourelly999: Addressed

src/v2i-hub/CDASimAdapter/src/CDASimConnection.cpp
dan-du-car: I get the optional part. I am trying to make a note that whether we need to set ` message["sensors"] = sensors_json_v;` when it is empty because I see it is not set when the path is empty.

configuration/README.md
```diff
@@ -44,14 +44,17 @@ Enter the login credentials you created in step 5b and login.
...
+
+* **SIMULATION_MODE** – Environment variable for enabling simulation components for V2X-Hub. If set to "true" or "TRUE" simulation components will be enable. Otherwise, simulation components will not be enabled.
+* **KAFKA_BROKER_ADDRESS** – Environment variable for storing Kafka broker connection string (including port).
+* **TIME_SYNC_TOPIC** – Environment variable for storing Kafka time sync topic.
+* **SIMULATION_IP** – Environment variable for storing IP address of CDASim application.
+* **SIMULATION_REGISTRATION_PORT** – Environment variable for storing port on CDASim that handles registration attempts.
+* **TIME_SYNC_PORT** – Environment varaible for storing port for receiving time sync messages from CDASim.
+* **V2X_PORT** – Environment variable for storing port for receiving v2x messages from CDASim
+* **SIM_V2X_PORT** – Environment variable for storing port for sending v2x messages to CDASim
+* **LOCAL_IP** – Environment variable for storing local IP of V2X Hub.
```
DokurOmkar, line 58: Currently we are passing the "LOCAL_IP" as "V2XHUB_IP" from .env file. Not sure if we need to update all the files to use "V2XHUB_IP" everywhere.

paulbourelly999: Will update to V2X-Hub IP

src/v2i-hub/CDASimAdapter/src/CDASimConnection.cpp
paulbourelly999: We do not need the that part when the sensor configuration file is empty or missing

DokurOmkar (approved):

PR #623, opened 2024-07-11
VH 1324 - Filter the versions returned from the initialization script 

DokurOmkar:
# PR Details
## Description

In this PR, updated the docker compose file to change the "LOCAL_IP" variable to "V2XHUB_IP" and updated the initialization script to remove the following from the returned versions:
- Curly braces: **^{}**
- Properties found 
- Duplicate entries

## How Has This Been Tested?
This has been tested by running the script locally. A screenshot is attached below after testing.
Screenshot from 2024-07-11 08-42-58

## Types of changes

## Checklist:


configuration/initialization.sh
```diff
@@ -23,9 +23,12 @@ latest_version=$(echo "$release_info" | grep -o '"tag_name": *"[^"]*"' | cut -d
 # Fetching all tags from Git repository
 tags=$(git ls-remote --tags https://github.com/usdot-fhwa-OPS/V2X-Hub.git | awk -F/ '{print $3}' | sort -V)
```
paulbourelly999, line 24: On issue I am realizing right now is that getting tags may return tags that are not associated with any release. Maybe we should considered install `gh' as part of this script and using the output of  `gh release list` which looks like this 
image

paulbourelly999 (changes requested):
Possible enhancement

configuration/initialization.sh
DokurOmkar: I think using "gh" would require an additional step of installing it. As part of this bug, we can work on removing the unwanted characters and add a follow up story to make an enhancement.

In the follow up story, we can add a logic to check if gh installed or not and add installation commands within this script.

paulbourelly999 (approved):

PR #627, opened 2024-07-31
Install carma-clock-1 from ubuntu distribution (correct version)

paulbourelly999:
# PR Details
## Description
This PR is to correct the version of `carma-clock-1` V2X-Hub pulls from our Debian repository. Previously it pulled from develop bucket which is no longer used and therefore contains versions of the library from 2023. Currently we push a debian package for each ubuntu distribution.

## Motivation and Context
Pull correct version of `carma-clock-1` package created from https://github.com/usdot-fhwa-stol/carma-time-lib

## How Has This Been Tested?
CI Process

## Types of changes

## Checklist:

### DIFF
--- a/scripts/install_dependencies.sh
+++ b/scripts/install_dependencies.sh
@@ -3,8 +3,11 @@
 # exit on errors
 set -e
 
+# Get ubuntu distribution code name. All STOL APT debian packages are pushed to S3 bucket based on distribution codename.
+. /etc/lsb-release
+
 # add the STOL APT repository
-echo "deb [trusted=yes] http://s3.amazonaws.com/stol-apt-repository develop main" > /etc/apt/sources.list.d/stol-apt-repository.list
+echo "deb [trusted=yes] http://s3.amazonaws.com/stol-apt-repository ${DISTRIB_CODENAME} main" > /etc/apt/sources.list.d/stol-apt-repository.list
 
 apt-get update
 


scripts/install_dependencies.sh
```diff
@@ -3,8 +3,11 @@
 # exit on errors
 set -e
 
+# Get ubuntu distribution code name. All STOL APT debian packages are pushed to S3 bucket based on distribution codename.
+. /etc/lsb-release
+
 # add the STOL APT repository
-echo "deb [trusted=yes] http://s3.amazonaws.com/stol-apt-repository develop main" > /etc/apt/sources.list.d/stol-apt-repository.list
+echo "deb [trusted=yes] http://s3.amazonaws.com/stol-apt-repository ${DISTRIB_CODENAME} main" > /etc/apt/sources.list.d/stol-apt-repository.list
```
dan-du-car, line 10: How to look at the list of available ubuntu distributions support of this carma-clock-1

paulbourelly999: Why would that be necessary. It is not necessary for any other package you install via apt?

dan-du-car: This package is limited and supported for ubuntu OS only? Also if the ubuntu is version 16.06 or 24.04, does this carma-clock-1 still exist? If not, do we have a visible document that tells users that this package is only compatible with certain ubuntu version?

paulbourelly999: I think this is excessive. When we release CARMA Platform or any other library we can document which ubuntu distributions we have built it on (https://github.com/usdot-fhwa-stol/carma-time-lib/pulls?tab=readme-ov-file#install-library) by listing prerequisites for using the application or library but this does not need to be available from the package manager. Additionally it is definitely not in the scope for this fix. When you install other applications via apt they do not list ubuntu compatibility. We have documented on carma-time-lib how to install this package via apt and provide an ubuntu distribution. Anything beyond that, I do not think is our responsibility.

dan-du-car (approved):

PR #649, opened 2024-11-12
Vh 1367:  update telemetry list json structure and serialization deserialization

dan-du-car:
# PR Details
## Description
This story is to update the Telemetry List (https://usdot-carma.atlassian.net/wiki/spaces/V2XH/pages/2688319550/V2xHub+UI+and+commandPlugin+message+exchanges#List.1 ) data JSON message structure and logic for serializing to and from JSON . Below is an outline of the current structure.

{
    "CARMACloud": {
        "id": "24",
        "description": "CARMA cloud plugin for making websocket connection with CARMA cloud .",
        "version": "7.6.0",
        "enabled": "Disabled",
        "path": "/var/www/plugins/CARMACloudPlugin",
        "exeName": "/bin/CARMACloudPlugin",
        "manifest": "manifest.json",
        "maxMessageInterval": "500000",
        "commandLineParameters": ""
    },
    "CARMAStreetsPlugin": {
        "id": "25",
        "description": "Plugin to communicate with CARMA Streets via Apache Kafka",
        "version": "7.6.0",
        "enabled": "Disabled",
        "path": "/var/www/plugins/CARMAStreetsPlugin",
        "exeName": "/bin/CARMAStreetsPlugin",
        "manifest": "manifest.json",
        "maxMessageInterval": "500000",
        "commandLineParameters": ""
    }
}

This message structure does not correspond well to the object data and therefore is more difficult to deserialize since attribute name information is dynamic and includes object data ( plugin name). Instead a JSON object defined as follows would require no custom processing for serialization into the correct object data, a array of plugin information.

 [   
    {
        "name": "CARMACloudPlugin"
        "id": "24",
        "description": "CARMA cloud plugin for making websocket connection with CARMA cloud .",
        "version": "7.6.0",
        "enabled": "Disabled",
        "path": "/var/www/plugins/CARMACloudPlugin",
        "exeName": "/bin/CARMACloudPlugin",
        "manifest": "manifest.json",
        "maxMessageInterval": "500000",
        "commandLineParameters": ""
    },
    {
        "name": "CARMAStreetsPlugin",
        "id": "25",
        "description": "Plugin to communicate with CARMA Streets via Apache Kafka",
        "version": "7.6.0",
        "enabled": "Disabled",
        "path": "/var/www/plugins/CARMAStreetsPlugin",
        "exeName": "/bin/CARMAStreetsPlugin",
        "manifest": "manifest.json",
        "maxMessageInterval": "500000",
        "commandLineParameters": ""
    }
]

This story includes scope for create data classes to store Telemetry List data as well as functions that use a JSON Library to serialize to and from JSON

## Related Issue
https://usdot-carma.atlassian.net/browse/VH-1367

## How Has This Been Tested?
Local integration test

## Types of changes

## Checklist:

### DIFF
--- a/src/tmx/TmxUtils/CMakeLists.txt
+++ b/src/tmx/TmxUtils/CMakeLists.txt
@@ -16,6 +16,7 @@ ENDIF ()
 
 TARGET_INCLUDE_DIRECTORIES (${PROJECT_NAME} SYSTEM PUBLIC
                             $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>
+                            $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/telemetry>
                             $<BUILD_INTERFACE:${TMXMESSAGES_INCLUDE_DIR}>
                             ${MYSQL_INCLUDE_DIR}
                             ${NETSNMP_INCLUDE_DIRS}
@@ -33,6 +34,7 @@ TARGET_LINK_LIBRARIES (${PROJECT_NAME} PUBLIC
                        rdkafka++ 
                        ::carma-clock
                        gmock
+                       jsoncpp
                        pthread m rt)
                        
 SET (TMXUTILS_LIBRARIES ${PROJECT_NAME} PARENT_SCOPE)
--- a/src/tmx/TmxUtils/src/telemetry/PluginTelemetry.cpp
+++ b/src/tmx/TmxUtils/src/telemetry/PluginTelemetry.cpp
@@ -1,28 +1,33 @@
 #include "PluginTelemetry.h"
 namespace tmx::utils::telemetry{
-    PluginInfo PluginTelemetry::getPluginInfo(){
+    PluginInfo PluginTelemetry::getPluginInfo() const{
         return _pluginInfo;
     }
 
-    PluginInstallation PluginTelemetry::getPluginInstallation(){
+    PluginInstallation PluginTelemetry::getPluginInstallation() const{
         return _pluginInstallation;
     }
 
     void PluginTelemetry::setPluginInfo(PluginInfo pluginInfo){
-        _isPluginInfoSet=true;
         _pluginInfo = pluginInfo;
     }
     
     void PluginTelemetry::setPluginInstallation(PluginInstallation pluginInstallation){
-        _isInstallationSet=true;
         _pluginInstallation = pluginInstallation;
     }
 
-    bool PluginTelemetry::isInstallationSet(){
-        return _isInstallationSet;
-    }
-
-    bool PluginTelemetry::isPluginInfoSet(){
-        return _isPluginInfoSet;
+    
+    ostream& operator <<(ostream& os, const PluginTelemetry& telemetry){
+        os << "id="<<telemetry._pluginInfo.id
+        <<",name="<<telemetry._pluginInfo.name
+        <<",description="<<telemetry._pluginInfo.description
+        <<",version="<<telemetry._pluginInfo.version
+        <<",enabled="<<telemetry._pluginInstallation.enabled
+        <<",exeName="<<telemetry._pluginInstallation.exeName
+        <<",manifest="<<telemetry._pluginInstallation.manifest
+        <<",path="<<telemetry._pluginInstallation.path
+        <<",commandLineParameters="<<telemetry._pluginInstallation.commandLineParameters
+        <<",maxMessageInterval="<<telemetry._pluginInstallation.maxMessageInterval;
+        return os;
     }
 }
--- a/src/tmx/TmxUtils/src/telemetry/PluginTelemetry.cpp
+++ b/src/tmx/TmxUtils/src/telemetry/PluginTelemetry.cpp
@@ -7,4 +7,22 @@ namespace tmx::utils::telemetry{
     PluginInstallation PluginTelemetry::getPluginInstallation(){
         return _pluginInstallation;
     }
+
+    void PluginTelemetry::setPluginInfo(PluginInfo pluginInfo){
+        _isPluginInfoSet=true;
+        _pluginInfo = pluginInfo;
+    }
+    
+    void PluginTelemetry::setPluginInstallation(PluginInstallation pluginInstallation){
+        _isInstallationSet=true;
+        _pluginInstallation = pluginInstallation;
+    }
+
+    bool PluginTelemetry::isInstallationSet(){
+        return _isInstallationSet;
+    }
+
+    bool PluginTelemetry::isPluginInfoSet(){
+        return _isPluginInfoSet;
+    }
 }
--- a/src/tmx/TmxUtils/src/telemetry/PluginTelemetry.cpp
+++ b/src/tmx/TmxUtils/src/telemetry/PluginTelemetry.cpp
@@ -1,0 +1,10 @@
+#include "PluginTelemetry.h"
+namespace tmx::utils::telemetry{
+    PluginInfo PluginTelemetry::getPluginInfo(){
+        return _pluginInfo;
+    }
+
+    PluginInstallation PluginTelemetry::getPluginInstallation(){
+        return _pluginInstallation;
+    }
+}
--- a/src/tmx/TmxUtils/src/telemetry/PluginTelemetry.h
+++ b/src/tmx/TmxUtils/src/telemetry/PluginTelemetry.h
@@ -35,11 +35,16 @@ namespace tmx::utils::telemetry
     private:
         PluginInfo _pluginInfo;
         PluginInstallation _pluginInstallation;
+        bool _isInstallationSet;
+        bool _isPluginInfoSet;
     public:
         PluginTelemetry()=default;
-        PluginTelemetry(PluginInfo pluginInfo, PluginInstallation pluginInstallation):_pluginInfo(pluginInfo), _pluginInstallation(pluginInstallation){}
         PluginInfo getPluginInfo();
         PluginInstallation getPluginInstallation();
+        void setPluginInfo(PluginInfo pluginInfo);
+        void setPluginInstallation(PluginInstallation pluginInstallation);
+        bool isInstallationSet();
+        bool isPluginInfoSet();
         ~PluginTelemetry()=default;
     };
     
--- a/src/tmx/TmxUtils/src/telemetry/PluginTelemetry.h
+++ b/src/tmx/TmxUtils/src/telemetry/PluginTelemetry.h
@@ -1,0 +1,46 @@
+#pragma once
+#include <string>
+using namespace std;
+
+namespace tmx::utils::telemetry
+{
+    struct PluginInfo
+    {
+        string id;
+        //Brief name for a plugin
+        string name; 
+        //Detailed description of the plugin and its usage
+        string description;
+        //Latest version of the plugin
+        string version;
+    };
+
+    struct PluginInstallation
+    {  
+        //Whether the plugin is current enabled or disabled
+        string enabled;
+        //Where the plugin executable is located
+        string path;
+        //The plugin executable name
+        string exeName;
+        //The manifest file that contains some configuration parameters
+        string manifest;
+        //Message broadcast interval
+        string maxMessageInterval;
+        string commandLineParameters;
+    };
+
+    class PluginTelemetry
+    {
+    private:
+        PluginInfo _pluginInfo;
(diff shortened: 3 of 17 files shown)


src/tmx/TmxUtils/CMakeLists.txt
```diff
@@ -33,6 +34,7 @@ TARGET_LINK_LIBRARIES (${PROJECT_NAME} PUBLIC
                        rdkafka++ 
                        ::carma-clock
                        gmock
+                       jsoncpp
```
paulbourelly999, line 37: Is this still necessary. No right since we are using boost property tree.

dan-du-car: Addressed https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/649/commits/8ecd06c6f047aef1f29efa18e5d717212d19b3ea

src/tmx/TmxUtils/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ ENDIF ()
 
 TARGET_INCLUDE_DIRECTORIES (${PROJECT_NAME} SYSTEM PUBLIC
                             $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>
+                            $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/telemetry>
```
paulbourelly999, line 19: Why is this line necessary for the new telemetry sub-package but not for the existing kafka, simulation, database sub-pacakges

src/tmx/TmxUtils/src/telemetry/PluginTelemetry.cpp
```diff
@@ -0,0 +1,88 @@
...
+    }
+    
+    void PluginTelemetry::deserialize(const boost::property_tree::ptree& jsonContainer){
+        try{
+            if(jsonContainer.empty()){
+                throw tmx::TmxException("Cannot deserialize empty JSON!");
+            }
+
+            PluginInfo pluginInfo = {
+                jsonContainer.get<string>("id",""),
```
paulbourelly999, line 67: Does get throw an exception if the key is not found?

src/tmx/TmxUtils/src/telemetry/TelemetryDeserializer.cpp
```diff
@@ -0,0 +1,16 @@
+#include "TelemetryDeserializer.h"
+
+namespace tmx::utils::telemetry
+{
+    boost::property_tree::ptree TelemetryDeserializer::stringToJson( const string& jsonString){
```
paulbourelly999, line 5: This function would benefit from better naming. This function reads a JSON string into a boost property tree. A name like readJson or jsonToPtree makes more sense to me since I property tree is not JSON itself.

src/tmx/TmxUtils/src/telemetry/PluginTelemetry.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+        auto maxMessageInterval = _pluginInstallation.maxMessageInterval;
+        output.store(message_path_type("maxMessageInterval"), maxMessageInterval);
+
+        auto commandLineParameters = _pluginInstallation.commandLineParameters;
+        output.store(message_path_type("commandLineParameters"), commandLineParameters);
+        return output;
+    }
+    
+    void PluginTelemetry::deserialize(const boost::property_tree::ptree& jsonContainer){
```
paulbourelly999, line 60: Is the deserialize functionality only used for unit testing?

src/tmx/TmxUtils/src/telemetry/TelemetryDeserializer.cpp
paulbourelly999: even deserializeJson would make more sense to me. If you have a better alternative I would also be open to that.

src/tmx/TmxUtils/src/telemetry/PluginTelemetry.h
```diff
@@ -0,0 +1,56 @@
...
+        PluginTelemetry()=default;
+        PluginInfo getPluginInfo() const;
+        PluginInstallation getPluginInstallation() const;
+        void setPluginInfo(const PluginInfo& pluginInfo);
+        void setPluginInstallation(const PluginInstallation& pluginInstallation);
+         /***
+         * @brief Serialize PluginTelemetry object into JSON container (tmx::message_container_type)
+         * @return A JSON container of tmx::message_container_type
+         */
+        tmx::message_container_type serialize() const;
```
paulbourelly999, line 51: I think the naming for this method is also misleading. Serialize implies we are taking a data structure and converting it into a format for sending or storing like json or bytes. In this case we are really only converting this data type to and from property tree. I think we could name this in that fashion `toPtree` or `fromPtree`. Another option would be to use naming convention similar to what other data structures in v2xhub use that use property trees. I think they use methods `set_contents` and `get_contents` to set the object based on a property tree or get the property tree of the data in a object,

src/tmx/TmxUtils/src/telemetry/TelemetryDeserializer.h
```diff
@@ -0,0 +1,55 @@
...
+    {
+    public:
+        TelemetryDeserializer() = default;
+        /***
+         * @brief Deserialize JSON string into a vector of telemetry objects
+         * @param JSON string object 
+         * @return Vector of telemetry objects
+         */
+        template <typename T>
+        static vector<T> desrializeFullTelemetryPayload(const boost::property_tree::ptree& jsonContainer){
```
paulbourelly999, line 25: method signature is misspelled `deserializeFullTelemetryPayload`

src/tmx/TmxUtils/src/telemetry/PluginTelemetry.cpp
dan-du-car: If `id` not found, it returns the second arguments which is default "".

src/tmx/TmxUtils/src/telemetry/TelemetryDeserializer.h
paulbourelly999, line 25: Is this also only used for unit testing

src/tmx/TmxUtils/src/telemetry/TelemetryHeader.h
```diff
@@ -0,0 +1,13 @@
...
+#include <string>
+
+using namespace std;
+namespace tmx::utils::telemetry{
+    struct TelemetryHeader
+    {
+        string type;
+        string subtype;
+        string encoding;
+        uint64_t timestamp;
```
paulbourelly999, line 11: why is flags not included as part of the header

src/tmx/TmxUtils/src/telemetry/TelemetrySerializer.cpp
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    tmx::message_container_type TelemetrySerializer::serializeTelemetryHeader(const TelemetryHeader& header){
+        tmx::message_container_type headerContainer;
+        boost::property_tree::ptree element;
+        element.push_back(boost::property_tree::ptree::value_type(TYPE_STRING, header.type));
+        element.push_back(boost::property_tree::ptree::value_type(SUBTYPE_STRING, header.subtype));
+        element.push_back(boost::property_tree::ptree::value_type(ENCODING_STRING, header.encoding));
+        element.push_back(boost::property_tree::ptree::value_type(TIMESTAMP_STRING,to_string(header.timestamp)));
+        element.push_back(boost::property_tree::ptree::value_type(FLAGS_STRING,"0"));
```
paulbourelly999, line 34: why is this hard coded and not pulled from header data type

src/tmx/TmxUtils/src/telemetry/PluginTelemetry.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+        auto description = _pluginInfo.description;
+        output.store(message_path_type("description"), description); 
+
+        auto version = _pluginInfo.version;
+            output.store(message_path_type("version"), version);
+
+        auto enabled = _pluginInstallation.enabled;
+        //enabled value -> 0: disabled, 1 enabled
+        output.store(message_path_type("enabled"), !enabled ? "Disabled": enabled > 0 ? "Enabled" : "External");
```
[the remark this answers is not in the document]
dan-du-car: image

src/tmx/TmxUtils/src/telemetry/TelemetryDeserializer.h
dan-du-car: yes

src/tmx/TmxUtils/src/telemetry/TelemetrySerializer.h
```diff
@@ -0,0 +1,68 @@
...
+
+#include <tmx/messages/message.hpp>
+
+#include "PluginTelemetry.h"
+#include "TelemetrySerializerException.h"
+#include "TelemetryHeader.h"
+#include "TelemetryMetadata.h"
+
+namespace tmx::utils::telemetry{
+    class TelemetrySerializer
```
paulbourelly999, line 15: Same comment as before about using the serializer terminology. As I think about this, is this not more of a data type in that it is a collection of Telemetry objects? Then this class could be pretty synonymous with the PluginTelemetry class just as a vector of objects right?

paulbourelly999: I guess this also puts the information in a message_container

src/tmx/TmxUtils/test/TelemetrySerializerTest.cpp
```diff
@@ -0,0 +1,139 @@
...
+            telemetryPointer = make_shared<PluginTelemetry>();
+            telemetryPluginInfoOnlyPointer = make_shared<PluginTelemetry>();
+            telemetryPointer->setPluginInfo(pluginInfo);
+            telemetryPluginInfoOnlyPointer->setPluginInfo(pluginInfo);
+            telemetryPointer->setPluginInstallation(installation);
+        }
+        void TearDown() override{}
+    };
+
+    TEST_F(TelemetrySerializerTest, getPluginInfo){
```
paulbourelly999, line 50: NIT: This test case and the one below do not test the serializer but test the PluginTelemetry class. This should probably be put into its own unit test class

src/v2i-hub/CommandPlugin/src/CommandControl.cpp
```diff
@@ -129,114 +130,64 @@ void CommandPlugin::GetTelemetry(const string & dataType)
...
-			boost::algorithm::trim(pJSON);
-			//compare json, if not found or different add to updates
-			bool found = false;
-			if (pluginsJSON->find(pluginName) != pluginsJSON->end())
-				found = true;
-			if (!found || (*pluginsJSON)[pluginName] != pJSON)
+		auto payloadExist = output->get_storage().get_tree().get_child_optional("payload");
+		if(payloadExist){
+			auto payload = payloadExist.get();
+			BOOST_FOREACH(ptree::value_type &payloadItem, payload)
```
paulbourelly999, line 143: NIT: We should probably be removing BOOST_FOREACH calls and replacing them with C++ for each calls

src/tmx/TmxUtils/CMakeLists.txt
paulbourelly999: I think that is not the case. I think https://github.com/usdot-fhwa-OPS/V2X-Hub/blob/455ed7d9e4d178649e75dde1db4b59cea474ac70/src/tmx/TmxUtils/CMakeLists.txt#L2
```cmake
FILE (GLOB_RECURSE HEADERS "src/" "*.h*")
```
recursively goes through all the subdirectories so I don't think this is necessary at all. Looking at the documentation here https://cmake.org/cmake/help/v3.7/command/file.html for GLOB_RECURSE it seems that it "will traverse all the subdirectories of the matched directory and match the files."

paulbourelly999: When you remove this line what error do you get?

dan-du-car: Cannot compile because cannot find that header file

dan-du-car:
@paulbourelly999  Addressed your comments: https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/649/commits/e9db99c103cccaad1b0535e62f92fc84572ee2f8

PR #652, opened 2024-11-15
Feature 1218

paulbourelly999:
# PR Details
## Description
This PR updates the RSU Health Monitoring Plugin to support both the RSU 4.1 spec as well as the NTCIP1218 spec configurably. As part of the necessary changes, additions where may to the SNMP client to support SNMPV3 Authentication for NTCIP1218 with a Yunex RSU.

## Motivation and Context
Support new RSU standard (NTCIP1218)

## How Has This Been Tested?
Integration testing with Yunex RSU on the SDN 

## Types of changes

## Checklist:


src/build.sh
```diff
@@ -37,16 +37,12 @@ elif [ $RELEASE_BUILD -eq 0 ]; then
 fi
 pushd tmx
 cmake -Bbuild -DCMAKE_PREFIX_PATH=\"/usr/local/share/tmx\;\/opt/carma/cmake\;\" -DCMAKE_CXX_FLAGS="${COVERAGE_FLAGS}" -DCMAKE_C_FLAGS="${COVERAGE_FLAGS}" -DCMAKE_BUILD_TYPE="${BUILD_TYPE}" .
-pushd build
-make -j${numCPU}
-make install
-popd
+cmake --build build
```
dan-du-car, line 40: Shall we be consistent with the syntax either all cmake -Bbuild or cmake --build build

src/tmx/TmxUtils/src/SNMPClient.cpp
```diff
@@ -97,6 +96,156 @@ namespace tmx::utils
...
+        else if (securityLevel == "authNoPriv")
+        {
+            session.securityLevel = SNMP_SEC_LEVEL_AUTHNOPRIV;
+        }
+
+        else
+            session.securityLevel = SNMP_SEC_LEVEL_NOAUTH;
+
+        // Defining and generating authentication config
+        oid *usmAuthProto;
```
dan-du-car, line 143: Is it possible to put this in a separate function?

src/tmx/TmxUtils/src/SNMPClient.cpp
```diff
@@ -131,15 +276,14 @@ namespace tmx::utils
         // net-snmp has several methods for creating an OID object
         // their documentation suggests using get_node. read_objid seems like a simpler approach
         // TO DO: investigate update to get_node
-        if (!read_objid(input_oid.c_str(), OID, &OID_len))
-        {
-            // If oid cannot be created
+        if (!snmp_parse_oid(input_oid.c_str(), OID, &OID_len)) {
```
dan-du-car, line 279: It seems that they are similar, why we have a preference to this 
snmp_parse_oid method? comment still says read_objid

src/tmx/TmxUtils/src/SNMPClient.h
```diff
@@ -46,16 +46,24 @@ namespace tmx::utils
...
+        struct snmp_pdu *pdu;
+        /*Structure to hold response from the remote host*/
+        struct snmp_pdu *response;
 
         /*OID is going to hold the location of the information which we want to receive. It will need a size as well*/
         oid OID[MAX_OID_LEN];
         size_t OID_len = MAX_OID_LEN;
 
+        // Declare missing OID len definitions
+        #define USM_PRIV_PROTO_AES192_LEN 9
```
dan-du-car, line 62: Could we put this to the top of the file?

src/v2i-hub/RSUHealthMonitorPlugin/manifest.json
```diff
@@ -19,7 +19,7 @@
         },
         {
             "key":"RSUConfigurationList",
```
dan-du-car, line 21: Can we add the list of all possible authProtocol and privacyProtocol in the description for users?

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUConfigurationList.h
```diff
@@ -25,12 +28,17 @@ namespace RSUHealthMonitor
         RSUMIB_V_1218 = 2
     };
 
+    std::string mib_version_to_string(const RSUMibVersion &mibversion);
```
dan-du-car, line 31: I think we can make this function const

src/build.sh
paulbourelly999: Good call

src/v2i-hub/RSUHealthMonitorPlugin/src/RSUConfigurationList.h
paulbourelly999: It is a free function meaning it is not associated with any class or object so it is already const. I don't think I can even add that classifier to the function since it would not make sense. In a member function const before would indicate that the function does not modify the class/object it is a member of but in this case it is not a member of any class/object. That is my understanding.

src/v2i-hub/RSUHealthMonitorPlugin/manifest.json
paulbourelly999: Good call

src/tmx/TmxUtils/src/SNMPClient.h
paulbourelly999: That is definitely a good idea

paulbourelly999: @jwillmartin Where are these missing from? Why did we need to define these?

src/tmx/TmxUtils/src/SNMPClient.cpp
paulbourelly999: What would be our justification for removing this from the constructor?

src/tmx/TmxUtils/src/SNMPClient.cpp
paulbourelly999: I don't think it matters very much which of these methods we use as long as it works as intended and does not have any significant performance impact so I think this implementation is probably fine.

src/tmx/TmxUtils/src/SNMPClient.h
jwillmartin: @paulbourelly999 These lengths weren't declared in the snmp library where other lengths were, so I added them in.

paulbourelly999: Upon further investigation I think this needs to be limited to the class private scope. I do not thing we need access to these outside of the SNMP class scope.

src/tmx/TmxUtils/src/SNMPClient.cpp
dan-du-car: It's about readability of this constructor. I am suggesting move this logic ,usmAuthProto assignment, to a separate function and call from the constructor.

dan-du-car (approved):

src/tmx/TmxUtils/src/SNMPClient.h
dan-du-car: This is the suggestion from sonar code smell: https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&pullRequest=652&id=usdot-fhwa-ops_V2X-Hub&open=AZOTp-UF_EH9KHLovh3Q
Do you mean that the macro is in private class scope if defined inside class, and would be in public scope if defined at the top of a header file?

dan-du-car: This is what I read from online:```
Scope Issues:
Macros don't respect C++ scoping rules. A macro defined inside a class is still visible outside of it, which can lead to unexpected behavior and naming collisions.
```

PR #653, opened 2024-11-18
	modified:   src/v2i-hub/ImmediateForwardPlugin/manifest.json

jwillmartin:
modified:   src/v2i-hub/ImmediateForwardPlugin/src/ImmediateForwardPlugin.cpp
	modified:   src/v2i-hub/ImmediateForwardPlugin/src/ImmediateForwardPlugin.h

# PR Details
## Description

Updated IFM plugin based on feedback from Plugfest testing and other users. There's a need to change the Active Message File contents that the IFM plugin generates. These additions allows users to configure the TxMode and Signature fields without changing other functionality to the plugin.

## Motivation and Context

Users need to be able to configure some fields in the Immediate Forward Active Message File, which were previously hard-coded. 

## How Has This Been Tested?

Built docker image locally and tested that inputs to the new configurations were inserted into the Active Message File as expected. 

## Types of changes

## Checklist:

### DIFF
--- a/src/v2i-hub/ImmediateForwardPlugin/manifest.json
+++ b/src/v2i-hub/ImmediateForwardPlugin/manifest.json
@@ -14,7 +14,7 @@
     },
     {
       "key": "Messages_Destination_1",
-      "default": "{ \"Messages\": [ { \"TmxType\": \"SPAT-P\", \"SendType\": \"SPAT\", \"PSID\": \"0x8002\", \"Channel\": \"183\" }, { \"TmxType\": \"MAP-P\", \"SendType\": \"MAP\", \"PSID\": \"0x8002\", \"Channel\": \"183\" }, { \"TmxType\": \"PSM-P\", \"SendType\": \"PSM\", \"PSID\": \"0x27\", \"Channel\": \"183\" } ,{ \"TmxType\": \"TMSG07-P\", \"SendType\": \"TMSG07\", \"PSID\": \"0x8002\", \"Channel\": \"183\" },{ \"TmxType\": \"TMSG03-P\", \"SendType\": \"TMSG03\", \"PSID\": \"0xBFEE\", \"Channel\": \"183\" },{ \"TmxType\": \"TMSG05-P\", \"SendType\": \"TMSG05\", \"PSID\": \"0x8003\", \"Channel\": \"183\" }, { \"TmxType\": \"SSM-P\", \"SendType\": \"SSM\", \"PSID\": \"0x8002\", \"Channel\": \"183\" },{ \"TmxType\": \"SDSM\", \"SendType\": \"SensorDataSharingMessage\", \"PSID\": \"0x8010\", \"Channel\": \"183\" }] }",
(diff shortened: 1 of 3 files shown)


src/v2i-hub/ImmediateForwardPlugin/manifest.json
```diff
@@ -52,10 +52,20 @@
...
     {
       "key": "signMessage",
       "default": "0",
-      "description": "0 if False and 1 if True indicating whether should sign outgoing J2735 messages."
+      "description": "0 if False and 1 if True, indicating whether Active File will indicate to sign outgoing J2735 messages."
+    },
+    {
+      "key": "EnableHSM",
+      "default": "0",
+      "description": "0 if False and 1 if True, indicating whether HSM is connected and will sign outgoing J2735 messages."
```
dan-du-car, line 68: Is True HSM connected and sign the messages? What does HSM stand for?

src/v2i-hub/ImmediateForwardPlugin/src/ImmediateForwardPlugin.h
```diff
@@ -62,10 +62,12 @@ class ImmediateForwardPlugin : public tmx::utils::PluginClient
...
 	std::map<std::string, int> _messageCountMap;
-	std::string signatureData; 
-	std::string url; 
-	std::string baseurl; 
-	unsigned int signState; 
+	std::string signatureData;
+	std::string url;
+	std::string baseurl;
+	std::string txMode;
+	unsigned int signState;
```
dan-du-car, line 69: Does we still need any signState if replaced by enableHSM?

src/v2i-hub/ImmediateForwardPlugin/manifest.json
jwillmartin: Yeah, true means V2X Hub would try to sign the messages using a Hardware Security Module (only worked with one specific module).

src/v2i-hub/ImmediateForwardPlugin/src/ImmediateForwardPlugin.h
jwillmartin: In most cases, we don't sign the message and an RSU has to sign it. The RSU will expect to receive the Active Message file with this field set to True. So, this allows us to select True without trying to sign the message.

dan-du-car: - Is this Active Message file field set by IMF enableHSM? Or ? 
- Are you saying that we enableHSM = true without sign the message and the signState could be 1?
I am not sure I understand the concept of the enableHSM and signState and if there are any relationships between them. I see this comment before the if (enableHSM=true): `/// if signing is Enabled, request signing with HSM`. It sounds like there are some relationship between them?

jwillmartin: The field in the Active Message file is set by signState.
If you want V2X Hub to sign messages, you have to set enableHSM to 1. You should also set signState to 1 if your RSU is expecting to receive a signed message or will be signing a message. In some cases, you'll want to set both to 1, but in most cases, you'll only want to set signState to 1.

dan-du-car: If we are in most cases and we have enabledHSM to false, signState = 1, is this signState field value 1 updated in the Active Message File? 

Is this code here populate the active message file or?
`
os << "Signature=" << (signState == 1 ? "True" : "False") << "\n" << "Encryption=False\n";
`

jwillmartin: Yep! That's where it's updated

dan-du-car: Thanks! but if enabledHSM = false, this Signature is not populated.

jwillmartin: The entry for Signature is independent from enabledHSM. It's outside of that if-else case and only relies on signState.

dan-du-car (approved):

PR #656, opened 2024-12-11
5gaa ode plugin udp socket  support

dan-du-car:
# PR Details
## Description
Update ODEForwardPlugin to send UDP packet to remote UDP server ([ODE code base](https://github.com/usdot-jpo-ode)) by updating the following:
- Add ODEForwardPlugin communication mode: KAFKA or UDP
- Add UDP Clients to support different types of J2735 message forwarding as ODE server accept different UDP port connection for different J2735 messages.
- Update ODEForwardPlugin to support MAP (UDP socket only), SPAT (Both UDP and KAFKA), TIM (UDP socket only), BSM (Both UDP and KAFKA) messages

## Related Issue
NA

## Motivation and Context
5GAA demo and integration with ODE

## How Has This Been Tested?
Unit test

## Types of changes

## Checklist:

### DIFF
--- a/src/v2i-hub/ODEForwardPlugin/CMakeLists.txt
+++ b/src/v2i-hub/ODEForwardPlugin/CMakeLists.txt
@@ -5,3 +5,18 @@ SET (TMX_PLUGIN_NAME "ODEForwardPlugin")
 BuildTmxPlugin ( )
 
 TARGET_LINK_LIBRARIES (${PROJECT_NAME} tmxutils rdkafka++)
+
+#############
+## Testing ##
+#############
+enable_testing()
+include_directories(${PROJECT_SOURCE_DIR}/src)
+file(GLOB_RECURSE LIB_SOURCES LIST_DIRECTORIES false  src/*.cpp)
+add_library(${PROJECT_NAME}_lib ${LIB_SOURCES})
+target_link_libraries(${PROJECT_NAME}_lib PUBLIC tmxutils)
+set(BINARY ${PROJECT_NAME}_test)
+file(GLOB_RECURSE TEST_SOURCES LIST_DIRECTORIES false test/*.h test/*.cpp)
+set(SOURCES ${TEST_SOURCES} WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/test)
+add_executable(${BINARY} ${TEST_SOURCES})
+add_test(NAME ${BINARY} COMMAND ${BINARY})
+target_link_libraries(${BINARY} PUBLIC ${PROJECT_NAME}_lib gtest)
--- a/src/v2i-hub/ODEForwardPlugin/src/CommunicationModeHelper.h
+++ b/src/v2i-hub/ODEForwardPlugin/src/CommunicationModeHelper.h
@@ -49,13 +49,13 @@ namespace ODEForwardPlugin
          * @brief Get the Kafka mode
          * @return true if the communication mode is Kafka, false otherwise
          */
-        bool getKafkaMode();
+        bool getKafkaMode() const;
 
         /**
          * @brief Get the UDP mode
          * @return true if the communication mode is UDP, false otherwise
          **/
-        bool getUDPMode();
+        bool getUDPMode() const;
     };
     
 }
--- a/src/v2i-hub/ODEForwardPlugin/src/CommunicationModeHelper.h
+++ b/src/v2i-hub/ODEForwardPlugin/src/CommunicationModeHelper.h
@@ -13,15 +13,15 @@ namespace ODEForwardPlugin
      */
     enum class CommunicationMode{
         UDP,
-        KAFKA
+        KAFKA,
+        UNSUPPORTED
     };
 
     class CommunicationModeHelper
     {
     private:
-        //Indicator for the communication mode
-        bool _isKAFKAMode = false;
-        bool _isUDPMode = false;
+        //Indicator for the communication mode and default value is UNSUPPORTED
+        CommunicationMode _currentMode = CommunicationMode::UNSUPPORTED;
 
         //Map for communication mode to string
         std::unordered_map<CommunicationMode, std::string> _communicationModeMap = {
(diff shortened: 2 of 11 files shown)


paulbourelly999 (changes requested):
Is there a reason we are using UDP instead of Kafka. Last we spoke with Tony Kafka was still supported by ODE

dan-du-car:
> Is there a reason we are using UDP instead of Kafka. Last we spoke with Tony Kafka was still supported by ODE

@paulbourelly999  Hope you are in the email chain that @jwillmartin  has confirmed with Neaera team that UDP is the preferred method.

jwillmartin (approved):

src/v2i-hub/ODEForwardPlugin/src/ODEForwardPlugin.cpp
```diff
@@ -34,9 +34,12 @@
 
 	AddMessageFilter < BsmMessage > (this, &ODEForwardPlugin::HandleRealTimePublish);
  	AddMessageFilter < SpatMessage > (this, &ODEForwardPlugin::HandleSPaTPublish);
+	AddMessageFilter < MapDataMessage > (this, &ODEForwardPlugin::HandleMapPublish);
```
paulbourelly999, line 37: In the future we may want to also include a configuration parameter that sets which messages we forward. Seems weird to be able to configure the communication protocol but not messages but probably not a blocker for the planned demo

src/v2i-hub/ODEForwardPlugin/src/CommunicationModeHelper.h
```diff
@@ -0,0 +1,61 @@
...
+         * @brief Set the mode (UDP or KAFKA) of the communication
+         * @param modeSource The communication mode to be set
+         * @throws TmxException if the communication mode is unknown
+         */
+        void setMode(std::string& modeSource);
+        /**
+         * @brief Get the Kafka mode
+         * @return true if the communication mode is Kafka, false otherwise
+         */
+        bool getKafkaMode() const;
```
paulbourelly999, line 52: There is only one configuration parameter but the way this is implemented seems to suggest that both code be true simultaneously. Is there a reason why we made 2 boolean variables instead of 1 enumeration?

src/v2i-hub/ODEForwardPlugin/src/ODEForwardPlugin.cpp
```diff
@@ -137,28 +165,65 @@
...
+	}
+ }
+
+
+ void ODEForwardPlugin::HandleMapPublish(MapDataMessage &msg, routeable_message &routeableMsg) {
+	if(_communicationModeHelper->getUDPMode()){
+		sendUDPMessage(routeableMsg, UDPMessageType::MAP);
+	}else{
+		PLOG(logERROR) << "Unknown communication mode: " << _communicationMode;
+	}
```
paulbourelly999, line 202: This log statement is maybe confusing. Maybe we should updated to unsupported communication mode for message

src/v2i-hub/ODEForwardPlugin/src/CommunicationModeHelper.h
dan-du-car: Yeah, enum should work too.

dan-du-car: https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/656/commits/90a1d5cb362264e3b490242a73f163fda3a695b7

src/v2i-hub/ODEForwardPlugin/CMakeLists.txt
```diff
@@ -5,3 +5,18 @@ SET (TMX_PLUGIN_NAME "ODEForwardPlugin")
 BuildTmxPlugin ( )
 
 TARGET_LINK_LIBRARIES (${PROJECT_NAME} tmxutils rdkafka++)
+
+#############
+## Testing ##
+#############
+enable_testing()
+include_directories(${PROJECT_SOURCE_DIR}/src)
```
paulbourelly999, line 13: NIT: instead of using global commands in CMake that modify path for all targets, it is normally better to use target based commands like target_include_directories .

src/v2i-hub/ODEForwardPlugin/src/UDPMessageForwarder.cpp
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    std::string UDPMessageForwarder::hexToBytes(const std::string& hex) const{
+        std::vector<unsigned char> bytes;
+        for (size_t i = 0; i < hex.length(); i += 2) {
+            std::string byteString = hex.substr(i, 2);
+            unsigned char byte = static_cast<unsigned char>(std::stoul(byteString, nullptr, 16));
+            bytes.push_back(byte);
+        }
+        std::cout << std::endl;
```
paulbourelly999, line 49: What is the purpose of this statement

src/v2i-hub/ODEForwardPlugin/src/UDPMessageForwarder.h
```diff
@@ -0,0 +1,63 @@
...
+    class UDPMessageForwarder
+    {
+    private:
+        //Map for udp client and udp message type
+        std::unordered_map<UDPMessageType, std::shared_ptr<UdpClient>> _udpClientsMap;
+    public:
+        explicit UDPMessageForwarder() = default;
+        ~UDPMessageForwarder() = default;
+        /**
+         * @brief Send difference type of messages to via their corresponding udp client
```
paulbourelly999, line 32: NIT: Type "Send different types of messages to their corresponding udp clients"

src/v2i-hub/ODEForwardPlugin/test/main.cpp
paulbourelly999, line 1: NIT: I don't think we need main CPP for gtest. I think you can simply link `gtest_main` instead of 'gtest'. Address at your own discretion but it is probably best for us to reduce the amount of unnecessary code we write where possible. Here is some documentation from gtest https://google.github.io/googletest/primer.html#writing-the-main-function that explains than most users "should not need to write their own main function and instead link with gtest_main (as opposed to with gtest), which defines a suitable entry point. See the end of this section for details"

paulbourelly999 (approved):

PR #657, opened 2024-12-11
Feature/ped plugin additions

jwillmartin:
# PR Details
## Description
Made changes to Pedestrian Plugin to:
- Convert incoming FLIR data to SDSM and TIM messages, simultaneously.
- Accept user XML inputs to encode PSM, SDSM, TIM messages.

## Motivation and Context
These changes are in support of the upcoming 5GAA demo at TFHRC. We decided to stop generating PSMs from the FLIR data as sensor data should not be used to generate PSMs. However, sensor data is appropriate for generating SDSMs and TIMs.

In order to keep some old functionality, users can opt to send their own XMLs for these three messages to the plugin for encoding. 

## How Has This Been Tested?
Tested locally using the scripts found under PedestrianPlugin/scripts/.
Tested at TFHRC and correctly generated TIMs and SDSMs from live FLIR data. Messages were transmitted to an RSU and decoded by a third-party application. Message contents for both TIM and SDSM appeared visually valid on map overlays.

## Types of changes

## Checklist:


scripts/deployment_dependencies.sh
```diff
@@ -21,6 +21,7 @@ DEPENDENCIES="build-essential \
     libssl-dev \
     libwebsockets-dev \
     libxerces-c-dev \
+    python3-websockets \
```
paulbourelly999, line 24: If this is a dependency of a testing or mocking script can we instead add a requirements.txt with this dependency where the script is. This would allow users of the script an easy way to install dependencies of the script if they need it but would prevent us from bloating our software with things that are not actually dependencies of our system

src/tmx/TmxUtils/test/J2735MessageTest.cpp
```diff
@@ -749,7 +760,7 @@ TEST_F(J2735MessageTest, EncodeTravelerInformation){
...
...th><anchor><lat>389549153</lat><long>-771488965</long><elevation>400</elevation></anchor><laneWidth>366</laneWidth><directionality><forward/></directionality><closedPath><false/></closedPath><direction>0000000000000000</direction><description><path><offset><xy><nodes><NodeXY><delta><node-LatLon><lon>-771489394</lon><lat>389549194</lat></node-LatLon></delta><attributes><dElevation>-10</dElevation></attributes></NodeXY><NodeXY><delta><node-LatLon><lon>-771487215</lon><lat>389548996</lat></node-LatLon></delta><attributes><dElevation>10</dElevation></attributes></NodeXY><NodeXY><delta><node-LatLon><lon>-771485210</lon><lat>389548981</lat></node-LatLon></delta><attributes><dElevation>10</dElevation></attributes></NodeXY></nodes></xy></offset></path></description></GeographicalPath></regions><notUsed2>0</notUsed2><notUsed3>0</notUsed3><content><speedLimit><SEQUENCE><item><itis>27</itis></item></SEQUENCE><SEQUENCE><item><text>Curve Ahead</text></item></SEQUENCE><SEQUENCE><item><itis>2564</itis></item></SEQUENCE><SEQUENCE><item><text>25</text></item></SEQUENCE><SEQUENCE><item><itis>8720</itis></item></SEQUENCE></speedLimit></content></TravelerDataFrame></dataFrames></TravelerInformation>";
```
paulbourelly999, line 752: Why did we change this unit test payload?

src/tmx/TmxUtils/test/J2735MessageTest.cpp
```diff
@@ -763,9 +774,19 @@ TEST_F(J2735MessageTest, EncodeTravelerInformation){
 	#if SAEJ2735_SPEC < 2020
 	string expectedHex = "001f526011c35d000000000023667bac0407299b9ef9e7a9b9408230dfffe4386ba00078005a53373df3cf5372810461b90ffff53373df3cf53728104618129800010704a04c7d7976ca3501872e1bb66ad19b2620";
 	#else
-	string expectedHex = "001f6820100000000000f9e1d8d0803299b9eac27a9baa74232000000fcec0a9df4028007e53373d584f53754e846400b720000000b8f5374e3666e7ac5013ece3d4ddc1099b9e988050538f5378f9666e7a5a814140034000dea1f5e5db2a083a32e1c80a048b26a22100";
+	string expectedHex = "001f45201000000000019b3915c0807299b9ea847a9b9dea2001fffe4fd0f23a000078005253373d508f5373bd4400325fffe05b94cdcf53e3d4dcef30e7aa58000002250e1a8880";
 	#endif
-	ASSERT_EQ(expectedHex, timEnc.get_payload_str());			
+	ASSERT_EQ(expectedHex, timEnc.get_payload_str());
```
paulbourelly999, line 779: NIT: For unit tests use 
```cpp
EXPECT_EQ(...)
// instead of 
ASSERT_EQ(...)
```

EXPECT_EQ:
Generates a non-fatal failure if the values are not equal.
The test continues to execute even after the failure, allowing you to potentially discover multiple failures within a single test case.
ASSERT_EQ:
Generates a fatal failure if the values are not equal.
The test execution aborts immediately, exiting the current function.

paulbourelly999: In most cases we want EXPECT_EQ. This allows us to see all the errors at once. ASSERT is normally used when you want execution to stop because some assertion has failed and makes all other down stream code invalid.

src/tmx/TmxUtils/test/J2735MessageTest.cpp
```diff
@@ -763,9 +774,19 @@ TEST_F(J2735MessageTest, EncodeTravelerInformation){
...
+
+	// Decode any TIM
+	std::string timInput = "001f45201000000000019b3915c0807299b9ea847a9b9dea2001fffe4fd0f23a000078005253373d508f5373bd4400325fffe05b94cdcf53e3d4dcef30e7aa58000002250e1a8880";
+    std::vector<uint8_t> byteArray = hexStringToByteArray(timInput);
+    tmx::messages::TimEncodedMessage decodedTim;
+    decodedTim.set_data(byteArray);
+    auto decodedTimPtr = decodedTim.decode_j2735_message();
+
+    // Print out the decoded message
+    std::cout << "Decoded TIM: " << decodedTimPtr << std::endl;
```
paulbourelly999, line 789: Please  remove any print statements

src/v2i-hub/PedestrianPlugin/scripts/websocketServer.py
```diff
@@ -0,0 +1,51 @@
...
+            await sendMessage(websocket)
+        except websockets.exceptions.ConnectionClosed as e:
+            print(f"Connection closed: {e}")
+    else:
+        print(f"Client attempted to connect to invalid path: {path}")
+        await websocket.close(code=4000, reason="Invalid path")
+
+async def main():
+    # Start server at port 9000, as port 80 is used by V2X Hub
+    server = await websockets.serve(handleClient, "127.0.0.1", 9000)
```
paulbourelly999, line 46: NIT: Use argparse to allow you to easily expose arguements and description and help text for your python script. https://github.com/usdot-fhwa-OPS/V2X-Hub/blob/4f5db03af87d88111a0a7360edea7aad5495d987/src/v2i-hub/MUSTSensorDriverPlugin/scripts/MockMUSTSensor.py#L67

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
```diff
@@ -6,10 +6,44 @@ using namespace boost::property_tree;
 
 namespace PedestrianPlugin
 {
+    int 
+    FLIRWebSockAsyncClnSession::calculateMinuteOfYear(int year, int month, int day, int hour, int minute, int second) {
```
paulbourelly999, line 10: NIT: why is the method declaration on a separate line from the return type. This looks weird.

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
```diff
@@ -6,10 +6,44 @@ using namespace boost::property_tree;
...
+    startOfYear.tm_mday = 1;
+    startOfYear.tm_hour = 0;
+    startOfYear.tm_min = 0;
+    startOfYear.tm_sec = 0;
+
+    // Calculate difference in seconds
+    std::time_t current = std::mktime(&currentTime);
+    std::time_t start = std::mktime(&startOfYear);
+    if (current == -1 || start == -1) {
+        throw std::runtime_error("Failed to compute time difference");
```
paulbourelly999, line 33: Can this error be more descriptive. Not sure what this means. What is failing and why?

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
```diff
@@ -6,10 +6,44 @@ using namespace boost::property_tree;
...
+    int minuteOfYear = secondsDifference / 60;
+
+    return minuteOfYear;
+    }
+
     void
     FLIRWebSockAsyncClnSession::fail(beast::error_code ec, char const* what) const
     {
-        PLOG(logDEBUG) << what << ": " << ec.message() << std::endl;
+        PLOG(logDEBUG) << what << ": " << ec.message();
```
paulbourelly999, line 46: If this is a error message maybe we should change the logging to error level

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
```diff
@@ -42,7 +76,7 @@ namespace PedestrianPlugin
          beast::error_code ec,
          tcp::resolver::results_type results)
     {
-        PLOG(logDEBUG) << "In FLIRWebSockAsyncClnSession::on_resolve " << std::endl;
+        PLOG(logDEBUG) << "In FLIRWebSockAsyncClnSession::on_resolve ";
```
paulbourelly999, line 79: Can we delete this log statement. Not a big fan of log statement that just mark where you are in the code. Alternatively providing some information about current state would be helpful

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
```diff
@@ -61,7 +95,7 @@ namespace PedestrianPlugin
     void
     FLIRWebSockAsyncClnSession::on_connect(beast::error_code ec, tcp::resolver::results_type::endpoint_type ep)
     {
-        PLOG(logDEBUG) << "In FLIRWebSockAsyncClnSession::on_connect " << std::endl;
+        PLOG(logDEBUG) << "In FLIRWebSockAsyncClnSession::on_connect ";
```
paulbourelly999, line 98: Same comment as above

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
```diff
@@ -99,7 +133,7 @@ namespace PedestrianPlugin
     void
     FLIRWebSockAsyncClnSession::on_handshake(beast::error_code ec)
     {
-        PLOG(logDEBUG) << "In FLIRWebSockAsyncClnSession::on_handshake " << std::endl;
+        PLOG(logDEBUG) << "In FLIRWebSockAsyncClnSession::on_handshake ";
```
paulbourelly999, line 136: Same comment as above

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
```diff
@@ -116,7 +150,7 @@ namespace PedestrianPlugin
         beast::error_code ec,
         std::size_t bytes_transferred)
     {
-        PLOG(logDEBUG) << "In FLIRWebSockAsyncClnSession::on_write " << std::endl;
+        PLOG(logDEBUG) << "In FLIRWebSockAsyncClnSession::on_write ";
```
paulbourelly999, line 153: Same comment as above

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
```diff
@@ -136,7 +170,7 @@ namespace PedestrianPlugin
         beast::error_code ec,
         std::size_t bytes_transferred)
     {
-        PLOG(logDEBUG) << "In FLIRWebSockAsyncClnSession::on_read " << std::endl;
+        PLOG(logDEBUG) << "In FLIRWebSockAsyncClnSession::on_read ";
```
paulbourelly999, line 173: Same comment as above

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
```diff
@@ -148,15 +182,15 @@ namespace PedestrianPlugin
 	    std::string text(beast::buffers_to_string(buffer_.data()));
         ss<<text; 
 
-        PLOG(logDEBUG) << "Received:  " << text.c_str() << std::endl;
+        PLOG(logDEBUG) << "Received:  " << text.c_str();
```
paulbourelly999, line 185: We should be able to just print text without converting it to a c string right?

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
```diff
@@ -226,6 +261,10 @@ namespace PedestrianPlugin
                             int str_length_diff = 8 - idResult.length();
                             idResult.append(str_length_diff, '0');
 
+                            // Need to convert ID for SDSM
+                            // Map the value to be less than 65535
+                            int mappedID = static_cast<int>(id % 65535);
```
paulbourelly999, line 266: What ID is this. I don't remember any none string or int IDs in the SDSM message.

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
```diff
@@ -255,46 +294,61 @@ namespace PedestrianPlugin
...
 
                         PLOG(logINFO) << "Received FLIR camera data for pedestrian " << idResult << " at location: (" << lat << ", " << lon <<
-                        ")" << ", travelling at speed: " << speed << ", with heading: " << alpha << " degrees" << std::endl;
+                        ")" << ", travelling at speed: " << speed << ", with heading: " << alpha << " degrees";
 
-                        PLOG(logINFO) << "PSM message count: " << msgCount << std::endl;
+                        PLOG(logINFO) << "Message count: " << msgCount;
 
-                        //constructing xml to send to BroadcastPSM function
+                        /* NOTE: Removed PSM constuction from FLIR input. Keeping as reference and if returned use is desired.
```
paulbourelly999, line 304: Can't you just use a configurable parameter to select what type of message you want to send. You can make the configurable parameter a comma separate list of approved types. Something like SDSM,TIM would trigger you to send and SDMS and TIM while something like PSM would just send a PSM

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
```diff
@@ -311,7 +365,7 @@ namespace PedestrianPlugin
     void
     FLIRWebSockAsyncClnSession::on_close(beast::error_code ec)
     {
-        PLOG(logDEBUG) << "In FLIRWebSockAsyncClnSession::on_close " << std::endl;
+        PLOG(logDEBUG) << "In FLIRWebSockAsyncClnSession::on_close ";
```
paulbourelly999, line 368: Same comment as earlier

src/v2i-hub/PedestrianPlugin/src/PedestrianPlugin.cpp
```diff
@@ -204,7 +204,7 @@ void PedestrianPlugin::UpdateConfigSettings()
     }
 	else
 	{
-		PLOG(logWARNING) << "Invalid configured data provider. Pedestrian Plugin requires valid data provider (FLIR, PSM)!";
+		PLOG(logWARNING) << "Invalid configured data provider. Pedestrian Plugin requires valid data provider (FLIR, PSM, SDSM)!";
```
paulbourelly999: Both PSM and SDSM are not valid dataprovider options in your logic

paulbourelly999 (dismissed):
I think you should add a configurable parameter for the types of message you want to broadcast. That way you do not have to comment out current functionality. Also I think the dataprovider options should be **REST** or **FLIR**

src/tmx/TmxUtils/test/J2735MessageTest.cpp
jwillmartin: Had to generate a different kind of message and check to make sure the XML formatting and values were correct

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
jwillmartin: There are two IDs in the SDSM - sourceID(octet string) and objectID(long). I'm taking the ID provided from the FLIR for the objectID and since the value has to be less than 65535, I'm running a modulo on it.

While we're on it, I'm also using the same info from the FLIR as the sourceID. Should we let the user set the sourceID or can we randomize the ID so that the source and object aren't using the same data? The incoming pedestrian ID from FLIR will change for every new pedestrian, so the source would end up changing and we might not want that.

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
jwillmartin: Was following the format that was made by someone else. I don't like it either. I'll change all the declarations

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
jwillmartin: This lets you know that no value was passed to the mktime function or it generated nothing in current and start. Updating the description

src/v2i-hub/PedestrianPlugin/src/PedestrianPlugin.cpp
jwillmartin: Caught and updated. Further updating to change to REST

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
jwillmartin: Agreed. Deleting all of them

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
jwillmartin: Was following the format that was made by someone else. I don't like it either. I'll change all the declarations

src/tmx/TmxUtils/test/J2735MessageTest.cpp
```diff
@@ -763,9 +774,19 @@ TEST_F(J2735MessageTest, EncodeTravelerInformation){
...
-	string expectedHex = "001f6820100000000000f9e1d8d0803299b9eac27a9baa74232000000fcec0a9df4028007e53373d584f53754e846400b720000000b8f5374e3666e7ac5013ece3d4ddc1099b9e988050538f5378f9666e7a5a814140034000dea1f5e5db2a083a32e1c80a048b26a22100";
+	string expectedHex = "001f45201000000000019b3915c0807299b9ea847a9b9dea2001fffe4fd0f23a000078005253373d508f5373bd4400325fffe05b94cdcf53e3d4dcef30e7aa58000002250e1a8880";
 	#endif
-	ASSERT_EQ(expectedHex, timEnc.get_payload_str());			
+	ASSERT_EQ(expectedHex, timEnc.get_payload_str());
+
+	// Decode any TIM
+	std::string timInput = "001f45201000000000019b3915c0807299b9ea847a9b9dea2001fffe4fd0f23a000078005253373d508f5373bd4400325fffe05b94cdcf53e3d4dcef30e7aa58000002250e1a8880";
+    std::vector<uint8_t> byteArray = hexStringToByteArray(timInput);
+    tmx::messages::TimEncodedMessage decodedTim;
```
dan-du-car, line 784: Should this is `encodedTim` rather than decodedTim?

src/tmx/TmxUtils/test/J2735MessageTest.cpp
```diff
@@ -763,9 +774,19 @@ TEST_F(J2735MessageTest, EncodeTravelerInformation){
...
 	#else
-	string expectedHex = "001f6820100000000000f9e1d8d0803299b9eac27a9baa74232000000fcec0a9df4028007e53373d584f53754e846400b720000000b8f5374e3666e7ac5013ece3d4ddc1099b9e988050538f5378f9666e7a5a814140034000dea1f5e5db2a083a32e1c80a048b26a22100";
+	string expectedHex = "001f45201000000000019b3915c0807299b9ea847a9b9dea2001fffe4fd0f23a000078005253373d508f5373bd4400325fffe05b94cdcf53e3d4dcef30e7aa58000002250e1a8880";
 	#endif
-	ASSERT_EQ(expectedHex, timEnc.get_payload_str());			
+	ASSERT_EQ(expectedHex, timEnc.get_payload_str());
+
+	// Decode any TIM
+	std::string timInput = "001f45201000000000019b3915c0807299b9ea847a9b9dea2001fffe4fd0f23a000078005253373d508f5373bd4400325fffe05b94cdcf53e3d4dcef30e7aa58000002250e1a8880";
+    std::vector<uint8_t> byteArray = hexStringToByteArray(timInput);
```
dan-du-car, line 783: I understand this is a helper function for a unit test. Could we have a expected result for this helper function as well? So it helps to understand what the output/result is like?

src/v2i-hub/PedestrianPlugin/scripts/websocketServer.py
```diff
@@ -0,0 +1,51 @@
+import asyncio
+import websockets
+import json
+
+# Sample Pededertian Presence Tracking Data
```
dan-du-car, line 5: Provide description for the purpose of this script, how the communication flows between the Pedestrian plugin and this websocket server?

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
```diff
@@ -6,10 +6,44 @@ using namespace boost::property_tree;
 
 namespace PedestrianPlugin
 {
+    int 
+    FLIRWebSockAsyncClnSession::calculateMinuteOfYear(int year, int month, int day, int hour, int minute, int second) {
+    // Set up current time
+    std::tm currentTime = {};
+    currentTime.tm_year = year - 1900;
```
dan-du-car, line 13: Provide brief comments for this calculation for year and month

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
```diff
@@ -6,10 +6,44 @@ using namespace boost::property_tree;
...
+    currentTime.tm_mday = day;
+    currentTime.tm_hour = hour;
+    currentTime.tm_min = minute;
+    currentTime.tm_sec = second;
+
+    // Set up start of the year
+    std::tm startOfYear = {};
+    startOfYear.tm_year = year - 1900;
+    startOfYear.tm_mon = 0;
+    startOfYear.tm_mday = 1;
```
dan-du-car, line 24: Provide comments for calculation/choice of init values of  the year, day , etc.

src/tmx/TmxUtils/test/J2735MessageTest.cpp
jwillmartin: It should be. Updating

src/tmx/TmxUtils/test/J2735MessageTest.cpp
jwillmartin: Good call, updating

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
paulbourelly999: I think the source ID is similar to the BSM ID and should be shared for all messages generated from V2X Hub. I think there may be other places were we write this ID. I would take a look at how we do it there. We probably just read it from a config parameter. Long term we want to pull this all from a single place like infrastructure ID

jwillmartin: Ok, for now I'm leaving it as being populated from the same FLIR pedestrian ID. We can look at using an Infrastructure ID or user input later.

src/v2i-hub/PedestrianPlugin/scripts/websocketServer.py
jwillmartin: Added description

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
jwillmartin: Added comments

src/v2i-hub/PedestrianPlugin/src/FLIRWebSockAsyncClnSession.cpp
jwillmartin: Added configurable parameter that accepts SDSM, TIM, PSM in any combination

src/v2i-hub/PedestrianPlugin/src/PedestrianPlugin.cpp
```diff
@@ -226,37 +247,104 @@ void PedestrianPlugin::OnStateChange(IvpPluginState state)
...
+        return "SensorDataSharingMessage";
+    }
+	else if (message.find("<TravelerInformation>") != std::string::npos)
+	{
+        return "TravelerInformation";
+    }
+	else return "Unknown Message Type";
+}
+
+void PedestrianPlugin::BroadcastPedDet(const std::string &msgJson)
```
dan-du-car, line 266: Shall we rename this as it now support multiple message types.

src/v2i-hub/PedestrianPlugin/src/PedestrianPlugin.cpp
```diff
@@ -226,37 +247,104 @@ void PedestrianPlugin::OnStateChange(IvpPluginState state)
...
+		timENC.encode_j2735_message(timMsg);
+
+		auto msg = std::make_unique<TimEncodedMessage>();
+		msg.reset();
+		msg.reset(dynamic_cast<TimEncodedMessage*>(factory.NewMessage(api::MSGSUBTYPE_TRAVELERINFORMATION_STRING)));
+		std::string enc = timENC.get_encoding();
+		msg->refresh_timestamp();
+		msg->set_payload(timENC.get_payload_str());
+		msg->set_encoding(enc);
+		msg->set_flags(IvpMsgFlags_RouteDSRC);
```
dan-du-car, line 336: Is this routed to RSU or just broadcast internally, then the Tim message is forwarded to ODE through ODEPlugin? 
image

dan-du-car (changes requested):

src/v2i-hub/PedestrianPlugin/src/PedestrianPlugin.cpp
jwillmartin: I renamed everything from PSM to a broader Ped. It's still only generating messages containing Pedestrian information, so I think this is good as is

src/v2i-hub/PedestrianPlugin/src/PedestrianPlugin.cpp
jwillmartin: It's broadcast internally then can be used by any plugin like ODE or Immediate Forward. I tested it with the Immediate Forward plugin and it worked fine

dan-du-car (approved):

src/v2i-hub/PedestrianPlugin/scripts/websocketServer.py
```diff
@@ -0,0 +1,60 @@
+## This script starts a mock WebSocket server for the Pedestrian Plugin to connect to. 
+## Once connected, the Ped Plugin will periodically receive FLIR's pedestrian presence tracking data.
+
+import asyncio
+import websockets
+import json
+import argparse
+
+# Sample Pededertian Presence Tracking Data
+message = {
```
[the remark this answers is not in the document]
jwillmartin: It's just a sample message to help a developer with testing. This is the exact format sent by FLIR. It just makes it easy to test without needing a FLIR camera to test with

maefromm (approved):

PR #660, opened 2024-12-23
Fix build and coverage scripts

paulbourelly999:
# PR Details
## Description
Fix build and coverage scripts to only create build and coverage infrmation when parameter is present

## Related Issue
NA

## Motivation and Context
Avoid unintentional debug or coverage builds. Debug and coverage builds slow the build process and result in executables that are slower and require more memory for debugging purposes. Debug and coverage builds also create many unnecessary files that bloat our docker images. This PR is intended to allow for more control by the user, when executing these scripts, whether they want to build executables with debug or coverage flags

## How Has This Been Tested?
Local builds

## Types of changes

## Checklist:


src/coverage.sh
paulbourelly999, line 1: A little unsure about this. This file is called coverage but really just executes all the tests. We could renaming it to test and leave the coverage parameter or we could remove the coverage parameter and just assume this script would only ever be executed when attempting to generate coverage metrics. One issue with that is that for local testing it would be nice to have a script that locally runs all of our tests.

.vscode/tasks.json
```diff
@@ -6,7 +6,7 @@
         {
             "label": "build",
```
dan-du-car, line 7: Does it make sense to change the label `build` to `build with debug` to reflect the debug option?

.vscode/tasks.json
```diff
@@ -34,7 +34,7 @@
         {
             "label": "test",
             "type": "shell",
-            "command": "cd src && ./install_coverage_dependencies.sh && ./coverage.sh",
+            "command": "cd src && ./test.sh",
```
dan-du-car, line 37: - Is this coverage.sh renamed as test.sh? 
- install_coverage_dependencies.sh is removed , but is this gvor installed with some other scripts?

paulbourelly999: Yes coverage is renamed as test.sh.  I removed from the test task in VSCode the functionality to generate coverage results by default since normally this is not the intention. Normally people running this task expect only that their unit tests are run. Instead they get may coverage files they do not expect or understand.

I can also add two different tasks. One task with just testing and one task with testing and coverage.

paulbourelly999: Added a test with coverage task

.vscode/tasks.json
paulbourelly999: Added a build with coverage tasks and a description to both

dan-du-car (approved):

PR #664, opened 2025-01-16
Cda1tenth plugin bsm

dan-du-car:
# PR Details
## Description
Implementing the BSM handler in the CDA 1Tenth Plugin, BSMs will be determined based on their type. Plugin logic will then convert the BSM to a JSON property tree form using the “toTree” function.

## Related Issue
NA

## Motivation and Context
CDA1Tenth Support

## How Has This Been Tested?
Unit test and local integration test

## Types of changes

## Checklist:

### DIFF
--- a/src/v2i-hub/CDA1TenthPlugin/CMakeLists.txt
+++ b/src/v2i-hub/CDA1TenthPlugin/CMakeLists.txt
@@ -10,7 +10,6 @@ find_package(Qt5Network REQUIRED)
 
 
 include_directories( "/usr/local/include/pdclient")
-include_directories(${PROJECT_SOURCE_DIR}/src)
 
 BuildTmxPlugin ( )
 
@@ -25,9 +24,10 @@ enable_testing()
 add_library(${PROJECT_NAME}_lib src/BSMConverter.cpp)
 target_link_libraries(${PROJECT_NAME}_lib PUBLIC 
                                         tmxutils)
+target_include_directories(${PROJECT_NAME}_lib PUBLIC 
+                                        ${PROJECT_SOURCE_DIR}/src)
 set(BINARY ${PROJECT_NAME}_test)
 file(GLOB_RECURSE TEST_SOURCES LIST_DIRECTORIES false test/*.h test/*.cpp)
-set(SOURCES ${TEST_SOURCES} WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/test)
 add_executable(${BINARY} ${TEST_SOURCES})
 add_test(NAME ${BINARY} COMMAND ${BINARY})
 target_link_libraries(${BINARY} PUBLIC ${PROJECT_NAME}_lib gtest)
--- a/src/v2i-hub/CDA1TenthPlugin/CMakeLists.txt
+++ b/src/v2i-hub/CDA1TenthPlugin/CMakeLists.txt
@@ -16,7 +16,7 @@ BuildTmxPlugin ( )
 
 target_link_libraries(${PROJECT_NAME} PRIVATE pdclient )
 target_link_libraries(${PROJECT_NAME} PRIVATE Qt5::Core Qt5::Network ssl crypto ) 
-target_link_libraries(${PROJECT_NAME} PUBLIC tmxutils jsoncpp)
+target_link_libraries(${PROJECT_NAME} PUBLIC tmxutils)
 
 #############
 ## Testing ##
--- a/src/v2i-hub/CDA1TenthPlugin/CMakeLists.txt
+++ b/src/v2i-hub/CDA1TenthPlugin/CMakeLists.txt
@@ -10,10 +10,25 @@ find_package(Qt5Network REQUIRED)
 
 
 include_directories( "/usr/local/include/pdclient")
+include_directories(${PROJECT_SOURCE_DIR}/src)
 
 BuildTmxPlugin ( )
 
 target_link_libraries(${PROJECT_NAME} PRIVATE pdclient )
 target_link_libraries(${PROJECT_NAME} PRIVATE Qt5::Core Qt5::Network ssl crypto ) 
-target_link_libraries(${PROJECT_NAME} PUBLIC tmxutils)
+target_link_libraries(${PROJECT_NAME} PUBLIC tmxutils jsoncpp)
+
+#############
+## Testing ##
+#############
+enable_testing()
+add_library(${PROJECT_NAME}_lib src/BSMConverter.cpp)
+target_link_libraries(${PROJECT_NAME}_lib PUBLIC 
+                                        tmxutils)
+set(BINARY ${PROJECT_NAME}_test)
+file(GLOB_RECURSE TEST_SOURCES LIST_DIRECTORIES false test/*.h test/*.cpp)
+set(SOURCES ${TEST_SOURCES} WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/test)
+add_executable(${BINARY} ${TEST_SOURCES})
+add_test(NAME ${BINARY} COMMAND ${BINARY})
+target_link_libraries(${BINARY} PUBLIC ${PROJECT_NAME}_lib gtest)
 
--- a/src/v2i-hub/CDA1TenthPlugin/src/BSMConverter.cpp
+++ b/src/v2i-hub/CDA1TenthPlugin/src/BSMConverter.cpp
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2024 LEIDOS.
+ * Copyright (C) 2025 LEIDOS.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this except in compliance with the License. You may obtain a copy of
--- a/src/v2i-hub/CDA1TenthPlugin/src/BSMConverter.cpp
+++ b/src/v2i-hub/CDA1TenthPlugin/src/BSMConverter.cpp
@@ -43,7 +43,7 @@ namespace CDA1TenthPlugin
     brakesChild.add_child("traction", tractionChild);
     // Add abs child
     pt absChild;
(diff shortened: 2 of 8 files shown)


john-chrosniak (commented):
This is fairly large PR so it would be helpful if you could make a video demoing the added functionality to provide more context (if that wouldn't be too much extra effort)

src/v2i-hub/CDA1TenthPlugin/src/CDA1TenthPlugin.cpp
```diff
@@ -15,587 +15,464 @@
...
+		// Database configuration
+		GetConfigValue<string>("Database_Username", _database_username);
+		GetConfigValue<string>("Database_Password", _database_password);
+		GetConfigValue<string>("Database_IP", _database_ip);
+		GetConfigValue<uint16_t>("Database_Port", _database_port);
+		GetConfigValue<string>("Database_Name", _database_name);
+		GetConfigValue<string>("BSM_Transmit_Topic", _transmitBSMTopic);
+
+		// TODO Update for new Web Service Configuration
+		uint16_t polling_frequency;
```
willjohnsonk, line 52: just a question, should we have one polling frequency or split it for the BSMs directly updating the UI vs the MOMs updating the SQL database?

I believe the default is around 5 seconds for the current Port Drayage UI since it's only updating a SQL database and not handling anything with rapid updates.

src/v2i-hub/CDA1TenthPlugin/src/CDA1TenthPlugin.cpp
```diff
@@ -15,587 +15,464 @@
...
+		// Host Port
+		GetConfigValue<uint16_t>("Webservice_Port", port);
+		// True for HTTPS
+		GetConfigValue<bool>("Webservice_Secure", secure);
+		// Polling Frequency in seconds
+		GetConfigValue<uint16_t>("Webservice_Polling_Frequency", polling_frequency);
+
+		// client = std::make_shared<WebServiceClient>(host, port, secure, polling_frequency);
+		// Port Holding Area Configurable location
+		GetConfigValue<double>("Holding_Lat", _holding_lat);
```
willjohnsonk, line 68: I think Holding will be depreciated with the new UI approach so we can probably remove this parameter (outside of BSM story)

src/v2i-hub/CDA1TenthPlugin/src/BSMConverter.cpp
willjohnsonk, line 1: Tagging @paulbourelly999 on this- In cases where a plugin is using standard behavior (in this case CDA 1Tenth using BSMs) do we have a better approach to reusing this type of code instead of potentially duplicating it in a plugin's src? Not in scope to address here but BSM conversion is definitely a reusable code asset across V2X Hub

src/v2i-hub/CDA1TenthPlugin/src/CDA1TenthPlugin.cpp
```diff
@@ -15,587 +15,464 @@
...
+			// Get next_action for given action_id
+			next_action_id = con->prepareStatement("SELECT next_action FROM freight WHERE action_id = ? ");
+			// Get current action for given action_id
+			current_action = con->prepareStatement("SELECT * FROM freight WHERE action_id = ? ");
+			// Get first action for vehicle
+			first_action = con->prepareStatement("SELECT * FROM first_action WHERE cmv_id = ? ");
+			// Insert action into freight table
+			insert_action = con->prepareStatement("INSERT INTO freight VALUES(?,?,?,?,?, UUID(), ?)");
+			// Get action_id of the previous action given action_id
+			get_action_id_for_previous_action = con->prepareStatement("SELECT action_id FROM freight WHERE next_action = ? and operation = ? ");
```
willjohnsonk, line 89: assuming all of these predefined SQL statements will be updated in a follow-up story/effort

willjohnsonk (commented):
Two general comments:
- @dan-du-car you mentioned that something was causing build errors, could you add an inline comment to that commented out section as "This causes build errors" so we know to address it?
- Do we follow any standard header practice for V2X Hub files? I saw authors and info on some, if we do want to keep doing that make sure to at least add yourself as the author on new files @dan-du-car 
 
Only did my first pass but we may want to approve to move ahead with the webservice changes in the next PR

src/v2i-hub/CDA1TenthPlugin/src/CDA1TenthPlugin.cpp
willjohnsonk: non critical for current PR

src/v2i-hub/CDA1TenthPlugin/src/CDA1TenthPlugin.cpp
willjohnsonk: non critical for current PR

src/v2i-hub/CDA1TenthPlugin/src/CDA1TenthPlugin.cpp
dan-du-car: The polling frequency is not used for BSMs as we are forwarding it at the same  frequency as the source (the vehicle).

src/v2i-hub/CDA1TenthPlugin/src/CDA1TenthPlugin.cpp
```diff
@@ -59,8 +60,8 @@ void CDA1TenthPlugin::UpdateConfigSettings() {
 	GetConfigValue<bool>("Webservice_Secure", secure);
 	// Polling Frequency in seconds
 	GetConfigValue<uint16_t>("Webservice_Polling_Frequency", polling_frequency);
-
-	client = std::make_shared<WebServiceClient>( host, port, secure, polling_frequency );
+	//TODO: Commented out due to compilation error
```
paulbourelly999, line 63: Why is there a compilation error when we attempt to send messages to web service? Please also explain why we do not plan to address this in the current story or Pull Request

src/v2i-hub/CDA1TenthPlugin/src/CDA1TenthPlugin.cpp
```diff
@@ -390,184 +395,50 @@ CDA1TenthPlugin::CDA1Tenth_Object CDA1TenthPlugin::readCDA1TenthJson( const ptre
...
-
-		std::stringstream traction;
-		traction << bsm->coreData.brakes.traction;
-		brakes["traction"] = traction.str();
-
-		uint8_t binary = bsm->coreData.brakes.wheelBrakes.buf[0] >> 3;
-		unsigned int brake_applied_status_type = 4;
-		// e.g. shift the binary right until it equals to 1 (0b00000001) to determine the location of the non-zero bit
-		for (int i = 0; i < 4; i ++)
+		catch (TmxException &ex)
```
paulbourelly999, line 436: Use const reference

src/v2i-hub/CDA1TenthPlugin/src/BSMConverter.cpp
paulbourelly999: I am not sure I understand the question here. @willjohnsonk @dan-du-car is this comment still relevant despite the committed changes. I only ask because this comment is on a line that looks like it no longer exists.

paulbourelly999 (commented):
Please add the CDA1Tenth plugin to the sonar scanner properties file to ensure its code is considered for linting and unit test coverage metrics

src/v2i-hub/CDA1TenthPlugin/CMakeLists.txt
```diff
@@ -10,10 +10,25 @@ find_package(Qt5Network REQUIRED)
 
 
 include_directories( "/usr/local/include/pdclient")
+include_directories(${PROJECT_SOURCE_DIR}/src)
```
paulbourelly999, line 13: Why is this line necessary? It should not be.

paulbourelly999 (changes requested):
If at all possible, we want to avoid merging PRs that include commenting out large sets of code.

src/v2i-hub/CDA1TenthPlugin/src/BSMConverter.cpp
willjohnsonk: Yes, referring to the existence of "BSMConverter.cpp" here. It specifically does generic BSM conversion in the source files of a plugin. I know we have talked about extracting common functionality like this out for reuse before on road user. Not sure how relevant that is here.

src/v2i-hub/CDA1TenthPlugin/src/BSMConverter.cpp
```diff
@@ -0,0 +1,225 @@
...
+      return "unknown";
+    }
+  }
+
+  string BSMConverter::toBrakeAppliedStatusString(const BrakeAppliedStatus_t &wheelBrakes)
+  {
+    std::string result;
+    for (int i = 0; i < 5; ++i)
+    {
+      result += (wheelBrakes.buf[i / 8] & (1 << (7 - (i % 8)))) ? '1' : '0';
```
john-chrosniak, line 199: Could you add a comment describing the logic of this line?

src/v2i-hub/CDA1TenthPlugin/src/BSMConverter.cpp
```diff
@@ -0,0 +1,225 @@
+/**
+ * Copyright (C) 2024 LEIDOS.
```
john-chrosniak, line 2: Nit: 2025 (happy new year!)

src/v2i-hub/CDA1TenthPlugin/src/BSMConverter.cpp
willjohnsonk: This is derived from j2735 representation spec and is used in several places, I don't think it adds a ton of value to try to explain it. We do not talk about these horrors. (It's doing a bitwise logic conversion based on how they encode certain attributes in the spec to keep message size down)

PR #687, opened 2025-03-10
Fix MAP Plugin not broadcasting MAP in simulation

paulbourelly999:
# PR Details
## Description
This Pull Request fixes a bug with the MAP plugin in which it only sends a single MAP messages in simulation mode. Through local integration testing in the Dev Container environment, it was discovered that which the programming manual claims that multiple calls to SubscribeMessages() should not overwrite message filters it does.

Since the PluginClockAwareClient calls attempts to subscribe to TimeSync messages in its constructor and the MapPlugin attempts to subscribe to ACT messages in it's constructor, the second call to Subscribe removes the first filter. This means the plugin no longer receives time sync messages and all calls to sleep on the carma clock will hang indefinitely. To address this we avoid calling subscribe in the base class and instead rely on the parent class to call subscribe. 

The bug behavior can also be caught by examining plugin state
image
If plugin state does not include **Simulation Time Step** when in simulation it can be concluded that the plugin message filter does not include the TimeSync message handling of the time sync message triggers updates to this status

## Motivation and Context
Fix time synchronization for MAP plugin and other impacted plugins

## How Has This Been Tested?
local integration testing and tested in CDASim deployment by @adev4a 

## Types of changes

## Checklist:

### DIFF
--- a/src/v2i-hub/CDASimAdapter/scripts/send_timestep_udp.py
+++ b/src/v2i-hub/CDASimAdapter/scripts/send_timestep_udp.py
@@ -11,11 +11,11 @@ import time
 #
 # TODO Move this script into a more permanent location
 count_num = 0
-
+sim_time =0
 def generate_time_sync():
     jsonResult = {
         "seq":count_num, 
-        "timestep":int(time.time()*1000)
+        "timestep":sim_time
     }
     jsonResult = json.dumps(jsonResult)
     return jsonResult
@@ -35,7 +35,8 @@ while True :
         count_num += 1
         sock.sendto(encoded_msg,host)
         print( encoded_msg.decode(encoding= 'UTF-8'), 'was sent to ', host)
-        time.sleep(5)
+        time.sleep(.1)
+        sim_time+= 100
     except socket.gaierror:
 
         print ('There an error resolving the host')
--- a/src/v2i-hub/CDASimAdapter/scripts/send_timestep_udp.py
+++ b/src/v2i-hub/CDASimAdapter/scripts/send_timestep_udp.py
@@ -9,9 +9,12 @@ import time
 # host and port. To Test the Time Sync functionality of the CDASimAdapter
 # set port to the value of the TIME_SYNC_PORT environment variable.
 #
-# TODO Move this script into a more permanent location
+# count_name used to populate the seq field in the time sync message
 count_num = 0
+# sim_time (ms) used to populate the timestep field in the time sync message
 sim_time =0
+# sleep_dur (s) used to control the rate at which time sync messages are sent
+sleep_dur = 0.1
 def generate_time_sync():
     jsonResult = {
         "seq":count_num, 
@@ -35,7 +38,7 @@ while True :
         count_num += 1
         sock.sendto(encoded_msg,host)
         print( encoded_msg.decode(encoding= 'UTF-8'), 'was sent to ', host)
-        time.sleep(.1)
+        time.sleep(sleep_dur)
         sim_time+= 100
     except socket.gaierror:
 
--- a/src/v2i-hub/MUSTSensorDriverPlugin/src/MUSTSensorDriverPlugin.cpp
+++ b/src/v2i-hub/MUSTSensorDriverPlugin/src/MUSTSensorDriverPlugin.cpp
@@ -22,6 +22,10 @@ namespace MUSTSensorDriverPlugin {
 	
 	MUSTSensorDriverPlugin::MUSTSensorDriverPlugin(const string &name): PluginClientClockAware(name)
 	{
+		if (isSimulationMode()) {
+			// Subscribe to the TimeSyncMessage
+			SubscribeToMessages();
+		}
 		mustSensorPacketReceiverThread = std::make_unique<tmx::utils::ThreadTimer>(std::chrono::milliseconds(5));
 	}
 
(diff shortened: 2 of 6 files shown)


src/v2i-hub/CDASimAdapter/scripts/send_timestep_udp.py
```diff
@@ -11,11 +11,11 @@
 #
 # TODO Move this script into a more permanent location
 count_num = 0
-
+sim_time =0
```
dan-du-car, line 14: What is the unit of measure for this sim_time?

src/v2i-hub/MUSTSensorDriverPlugin/src/MUSTSensorDriverPlugin.cpp
```diff
@@ -22,6 +22,10 @@ namespace MUSTSensorDriverPlugin {
 	
 	MUSTSensorDriverPlugin::MUSTSensorDriverPlugin(const string &name): PluginClientClockAware(name)
 	{
+		if (isSimulationMode()) {
+			// Subscribe to the TimeSyncMessage
+			SubscribeToMessages();
```
dan-du-car, line 27: If it is not in sim mode, does this must driver subscribe to message at all?

src/v2i-hub/SpatPlugin/src/SpatPlugin.cpp
```diff
@@ -23,6 +23,10 @@ using namespace tmx::utils;
 namespace SpatPlugin {
 
 	SpatPlugin::SpatPlugin(const std::string &name) :PluginClientClockAware(name) {
+		if (isSimulationMode()) {
+			// Subscribe to the TimeSyncMessage
+			SubscribeToMessages();
```
dan-du-car, line 28: The same comment as above.

src/v2i-hub/MUSTSensorDriverPlugin/src/MUSTSensorDriverPlugin.cpp
dan-du-car: Is there a way to check if there are filters exist, then subscribe whenever it detect at least one filter exist?

paulbourelly999: Not that I am aware of.  This condition only needs to exist for plugins that are integrated with simulation but do not by default subscribe to any message filters.

src/v2i-hub/SpatPlugin/src/SpatPlugin.cpp
paulbourelly999: No not on the SPAT plugin. See response above

src/v2i-hub/CDASimAdapter/scripts/send_timestep_udp.py
```diff
@@ -9,13 +9,16 @@
...
 # set port to the value of the TIME_SYNC_PORT environment variable.
 #
-# TODO Move this script into a more permanent location
+# count_name used to populate the seq field in the time sync message
 count_num = 0
-
+# sim_time (ms) used to populate the timestep field in the time sync message
+sim_time =0
+# sleep_dur (s) used to control the rate at which time sync messages are sent
+sleep_dur = 0.1
```
adev4a, line 17: nit: Its usually better for readability to include units in variable names, suggest renaming this variable instead of just adding in comment

src/v2i-hub/CDASimAdapter/scripts/send_timestep_udp.py
```diff
@@ -9,13 +9,16 @@
 # host and port. To Test the Time Sync functionality of the CDASimAdapter
 # set port to the value of the TIME_SYNC_PORT environment variable.
 #
-# TODO Move this script into a more permanent location
+# count_name used to populate the seq field in the time sync message
 count_num = 0
-
+# sim_time (ms) used to populate the timestep field in the time sync message
+sim_time =0
```
adev4a, line 15: nit: Can we call this sim_timestep_in_ms?

src/v2i-hub/MapPlugin/src/MapPlugin.cpp
```diff
@@ -8,6 +8,7 @@ namespace MapPlugin {
 		AddMessageFilter(IVPMSG_TYPE_SIGCONT, "ACT", IvpMsgFlags_None);
 		SubscribeToMessages();
 		errThrottle.set_Frequency(std::chrono::minutes(30));
+
```
adev4a, line 11: Is Subscribe to messages on Line 9 here calling the same subscription to TimeSync messages? This should also have a check for simulation mode

adev4a (commented):
It seems like most of the plugins in v2x hub call SubscribetoMessages(), if this function is only valid for simulation is there any reason we are not updating them all?

dan-du-car (approved):

src/v2i-hub/MapPlugin/src/MapPlugin.cpp
paulbourelly999: It will subscribe to all message filters added. This means that for plugins that are already adding message filters we can call subscribe at it will subscribe to both time sync and all other appropriate message filters. For plugins that do not add any message filters we have to check for simulation mode since attempting to call subscribe without any message filters will return an exception

PR #723, opened 2025-06-18
Add status information to FLIR Camera Driver Plugin

paulbourelly999:
…d Detection Modify count

# PR Details
## Description
This PR adds the following status counts to the FLIR Camera Driver Plugin
- Dropped Pedestrian detections
- Modified Pedestrian detections (special cases in which we modify FLIR Camera received detections due to confirmed bad or invalid data)

To do this we add a attribute to the SensorDetectedObject called isModified which we set to false by default, but is intended to be set to true when ever we need to make modifications to Sensor output due to bad or invalid data. 

In the case of the FLIR Camera, we occasionally receive detections that have no angle (heading) but have non-zero speed. In our testing this happens exclusively when objects become stationary. Since the SDSM detections require both speed and heading and since speed is not interpretable for perception with our heading, when this occurs we simply assume the object is stationary and drop speed to 0

## Motivation and Context
Prevent detection drops and provide additional diagnostic information in V2X Hub UI.

## How Has This Been Tested?
Unit Testing and local Integration testing.

## Types of changes

## Checklist:

### DIFF
--- a/src/v2i-hub/FLIRCameraDriverPlugin/src/FLIRCameraDriverPlugin.cpp
+++ b/src/v2i-hub/FLIRCameraDriverPlugin/src/FLIRCameraDriverPlugin.cpp
@@ -93,9 +93,18 @@ namespace FLIRCameraDriverPlugin
 				{
 					// Retrieve the message queue in each session and send each one to be broadcast, then pop.
 					std::queue<tmx::messages::SensorDetectedObject> currentMsgQueue = flirSession->getMsgQueue();
+					
+					droppedPedCount += flirSession->getDroppedPedCount();
+					SetStatus<uint>("Dropped Pedestrian Count", droppedPedCount);
+					flirSession->clearDroppedPedCount();
+
 					while(!currentMsgQueue.empty())
 					{		
 						auto message = currentMsgQueue.front();
+						if (message.get_isModified()) {
+							modifiedPedCount++;
+							SetStatus<uint>("Assumed Stationary Pedestrian Count", modifiedPedCount);
+						} 
 						PLOG(logDEBUG1) << "Sending Simulated SensorDetectedObject Message " << message;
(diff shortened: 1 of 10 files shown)


src/v2i-hub/FLIRCameraDriverPlugin/src/FLIRCameraDriverPlugin.cpp
```diff
@@ -91,17 +91,52 @@ namespace FLIRCameraDriverPlugin
...
+						
 						PLOG(logDEBUG1) << "Sending Simulated SensorDetectedObject Message " << message;
 						this->BroadcastMessage<tmx::messages::SensorDetectedObject>(message, _name, 0 , IvpMsgFlags_None);
+						if (message.get_isModified()) {
+							modifiedPedCount++;
+						}
 						currentMsgQueue.pop();
 					}
+					// Update total pedestrian detection count and clear the message queue.
+					totalPedCount += currentMsgQueue.size();
```
adev4a: This looks like it will always be empty. It should be called before the while loop if we want to use queue size.

src/v2i-hub/FLIRCameraDriverPlugin/src/FLIRCameraDriverPlugin.cpp
```diff
@@ -103,6 +103,8 @@ namespace FLIRCameraDriverPlugin
 					std::unordered_set<int> incomingPedestrianIds;
 					// Retrieve the message queue in each session and send each one to be broadcast, then pop.
 					std::queue<tmx::messages::SensorDetectedObject> currentMsgQueue = flirSession->getMsgQueue();
+					// Update total pedestrian detection count and clear the message queue.
```
adev4a: We can probably also update the comment here, since the message queue doesn't get cleared after the while loop

adev4a (approved):

PR #728, opened 2025-07-01
Add Memory Usage status to plugin base class

paulbourelly999:
# PR Details
## Description
Adding new **Memory Usage** status that pulls PSS for service as memory usage every 1 sec via the main method of the plugin. For plugins that override their main method, they will need to add calls to this protected method to gain this functionality. This should improve our ability to monitor V2X Hub Plugin Level performance and detect reliability issues like Memory leaks easier

## Related Issue
CDAD-155

## Motivation and Context
This should improve our ability to monitor V2X Hub Plugin Level performance and detect reliability issues like Memory leaks easier

## How Has This Been Tested?
Dev Container Screen shot of new status 
image

## Types of changes

## Checklist:

### DIFF
--- a/src/tmx/TmxUtils/src/PluginClient.cpp
+++ b/src/tmx/TmxUtils/src/PluginClient.cpp
@@ -15,6 +15,9 @@
 #include <sstream>
 #include <stdexcept>
 #include <thread>
+#include <fstream>
+#include <string>
+#include <regex>
 
 #include "PluginUtil.h"
 #include "PluginUpgrader.h"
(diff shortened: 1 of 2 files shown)


src/tmx/TmxUtils/src/PluginClient.cpp
```diff
@@ -335,6 +338,28 @@ void PluginClient::SetStartTimeStatus()
 	_isStartTimeStatusSet = true;
 }
 
+long PluginClient::getPss() const {
+	 long pss = 0;
+    std::ifstream smaps_file("/proc/self/smaps"); // Open the smaps file for the current process.
```
jwillmartin, line 343: When using "self" as the PID, does that only return the smaps file for one plugin or are other process smaps also printed out?

jwillmartin: Looks like many processes are returned. Can you add a way to filter for only the specific Plugin making this call?

jwillmartin: Looks like we get the sum of all of these PSS values. Looks good.

paulbourelly999: This should only return smaps for one plugin and add them up  as I understand it

paulbourelly999:
> Do we want to get the memory usage for all plugins or just the ones you made the changes to in this branch?

The challenge here is that this needs to be added to the main thread in order to get updates on some fixed period. The plugins I made changes to are ones that override the `PluginClient::Main` method but do not do anything special in there. Hence they did not need to override this method. I mentioned this in the PR description and in the added documentation but for Plugins that do do stuff in the main method we need to add this functionality and test it individually to make sure it does not interfere the existing override Main method functionality for the plugin

PR #734, opened 2025-07-03
Seperate initialization script into modular scripts for easy use

paulbourelly999:
# PR Details
## Description
Update initialization script to be more modular. Mostly just took existing script and separated it into parts. In the future we may only want to run subsets of this, especially when deploying with CARMA Streets. Separating this script into parts helps with that process.

## Related Issue
CDAD-150

## Motivation and Context
Improve deployment process to be more modular and customizable

## How Has This Been Tested?
Ran script locally to test deployment 

## Types of changes

## Checklist:


configuration/run_v2xhub.sh
```diff
@@ -0,0 +1,18 @@
+# #!/bin/bash
+echo "Running V2X Hub..."
+# Install necessary and useful apps
+sudo apt update -y 
+sudo apt-get install chromium-browser -y
```
willjohnsonk: is this script meant to be run every time v2x-hub goes up? having a line for update/installing chromium feels unnecessary. A conditional dpkg check for chromium browser may be better here

willjohnsonk (changes requested):

configuration/run_v2xhub.sh
paulbourelly999: Resolved

configuration/install_docker.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+echo "Installing Docker on Ubuntu..."
+# Install Docker on Ubuntu
+for pkg in docker.io docker-doc docker-compose podman-docker containerd runc; do sudo apt-get remove $pkg; done
```
jwillmartin, line 4: Do we have to run this remove every time? Would we be able to also check if these packages already exist?

paulbourelly999: Added it to the initialization script since that calls this script.

PR #751, opened 2025-08-06
Fix cmake linking gtest

paulbourelly999:
# PR Details
## Description
Due to incorrect CMake instructions, TmxUtils shared library was compiling its own test code into the shared library. This would have several negative impacts, one of which was that it would require some consumers of the library to include gtest libraries as dependencies. This seems to have been intermittent since other consumers where able to avoid including gtest libraries when consuming this library. It is unclear why the negative symptoms of this would only appear in some cases but that is likely why this bug has gone undetected.

In addition to fixing this bug I have updated all V2X Hub plugins to use the gtest_main library instead of defining their own custom main methods.

## Related Issue
NA

## Motivation and Context
Fix compilation

## How Has This Been Tested?
Unit Testing 

## Types of changes

## Checklist:


src/tmx/TmxUtils/CMakeLists.txt
```diff
@@ -1,13 +1,11 @@
-PROJECT ( tmxutils CXX )
-FILE (GLOB_RECURSE HEADERS "src/" "*.h*")
-FILE (GLOB_RECURSE SOURCES  "src/" "*.c*")
+project( tmxutils CXX )
+file(GLOB_RECURSE HEADERS "src/" "src/*.h*")
```
dan-du-car: Do we still need the first "src"? It seems to be ignored.

src/tmx/TmxUtils/CMakeLists.txt
```diff
@@ -24,16 +22,16 @@ TARGET_LINK_LIBRARIES (${PROJECT_NAME} PUBLIC
...
                        ${NETSNMPAGENT} 
                        ${NETSNMPMIBS} 
                        ${NETSNMP} 
                        ${NETSNMP_LIBRARIES} 
                        rdkafka++ 
                        ::carma-clock
-                       gmock
-                       pthread m rt)
+                       pthread
+                       m
```
dan-du-car: What are "m" and "rt"?

src/v2i-hub/CDASimAdapter/CMakeLists.txt
```diff
@@ -17,4 +17,4 @@ SET(SOURCES ${TEST_SOURCES} WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/test)
 ADD_EXECUTABLE(${BINARY} ${TEST_SOURCES})
 ADD_TEST(NAME ${BINARY} COMMAND ${BINARY})
 TARGET_INCLUDE_DIRECTORIES(${BINARY} PUBLIC /usr/local/lib src/) 
-TARGET_LINK_LIBRARIES(${BINARY} PUBLIC ${PROJECT_NAME}_lib gtest gmock tmxutils ::carma-clock jsoncpp) 
\ No newline at end of file
+TARGET_LINK_LIBRARIES(${BINARY} PUBLIC ${PROJECT_NAME}_lib gtest gmock gtest_main)
```
dan-du-car, line 20: Are carma-clock and jsoncpp no longer needed?

src/v2i-hub/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ SET (TMX_APPNAME tmx CACHE INTERNAL "Application name")
 SET (TMX_APPDESCR "V2I Hub" CACHE INTERNAL "Application description")
 LIST(APPEND CMAKE_PREFIX_PATH "/opt/carma/lib")
 LIST(APPEND CMAKE_MODULE_PATH "/opt/carma/cmake")
-
+LIST(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}")
```
dan-du-car, line 19: Are all the cmake modules in the CMAKE_SOURCE_DIR?

src/v2i-hub/CDASimAdapter/CMakeLists.txt
paulbourelly999: They are inherited from the library public dependencies

src/tmx/TmxUtils/CMakeLists.txt
paulbourelly999: Not sure

src/tmx/TmxUtils/CMakeLists.txt
paulbourelly999: Source is the relative path according to documentation. Not sure why it is being ignored but all file paths should be reported relative to this path. It is possible it expects an absolute path as the relative path parameter

src/v2i-hub/CMakeLists.txt
paulbourelly999: No Look above at line 18

src/tmx/TmxUtils/CMakeLists.txt
dan-du-car: "src/*.h*" shall already specify which folder to look at. Where in the documentation indicates that  additional "src" before wildcard is required?

src/v2i-hub/CMakeLists.txt
dan-du-car: Sorry for the confusion. I mean the additional customized cmake modules are all under ${CMAKE_SOURCE_DIR}? If there is no customized cmake modules under ${CMAKE_SOURCE_DIR} folder, do we still need line 19?

src/tmx/TmxUtils/CMakeLists.txt
paulbourelly999: src/.h is a regex pattern to find files. src is the relative path. Please read the following documentation (https://cmake.org/cmake/help/v3.7/command/file.html) 
image

src/v2i-hub/CMakeLists.txt
paulbourelly999: Line 19 checks for cmake configs installed via the v2i-hub or tmx CMakeLists. `/opt/carma/cmake` checks for apt installed cmake packages

src/tmx/TmxUtils/CMakeLists.txt
dan-du-car: image

Yeah, it seems `src/*.h*` shall be sufficient per the above example `/dir/*py`

paulbourelly999: Update the relative path argument only gets read if you specific the relative flag. Additionally it has no impact on the regex search and only impacts how file paths are reported. I have removed it. If we want to keep it the equivalent would look like this
```cmake
file(GLOB_RECURSE SOURCES RELATIVE ""${CMAKE_SOURCE_DIR}"/src" "src/*.c*")
```
This would return all the files as relative path from `/src/tmx/TmxUtils/src`

src/v2i-hub/CMakeLists.txt
dan-du-car: I see. Are cmake configs also installed in subdirectories of `${CMAKE_SOURCE_DIR} `folder? Does this `APPEND` search for all cmakes under the subdirectories?

dan-du-car (approved):

dan-du-car:
Looks good. Thanks!

PR #754, opened 2025-08-14
Add retry mechanism for setting rsumode for ntcip1218

paulbourelly999:
# PR Details
## Types of changes

## Checklist:

### DIFF
--- a/src/v2i-hub/ImmediateForwardPlugin/src/IMFNTCIP1218Worker.h
+++ b/src/v2i-hub/ImmediateForwardPlugin/src/IMFNTCIP1218Worker.h
@@ -39,6 +39,12 @@ namespace ImmediateForward {
      * @param client The SNMP client 
      */
     void setRSUMode(tmx::utils::snmp_client* const client, unsigned int mode  );
+
+    /**
+     * @brief Method that returns RSU Mode
+     * @param client tmx::utils::snmp_client
+     */
+    void waitForRSUModeStandby(tmx::utils::snmp_client* const client, unsigned int retry = 1 , unsigned int interval = 1 ) ;
     /**
      * @brief Initialize the immediate forward table on the RSU
      * @param client The SNMP client to use for the operation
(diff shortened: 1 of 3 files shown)


src/v2i-hub/ImmediateForwardPlugin/src/IMFNTCIP1218Worker.h
```diff
@@ -39,6 +39,12 @@ namespace ImmediateForward {
      * @param client The SNMP client 
      */
     void setRSUMode(tmx::utils::snmp_client* const client, unsigned int mode  );
+
+    /**
+     * @brief Method that returns RSU Mode
+     * @param client tmx::utils::snmp_client
+     */
+    void waitForRSUModeStandby(tmx::utils::snmp_client* const client, unsigned int retry = 1 , unsigned int interval = 1 ) ;
```
jwillmartin, line 47: Can you add the parameter description for retry and interval? Mostly want to specify that interval is in seconds

paulbourelly999: I kind of don't want to make this a configuration parameter since if we select reasonable values it will work for all RSUs and there will be no benefit as an end user for this configuration parameter. By reducing the amount of configuration parameters we make V2X Hub more usable and maintainable. Having this be a configuration parameter would just really introduce the possibility of an end-user misconfiguring this if we have reasonable values which should work for all RSUs. Currently we use 12 retries and 5 second interval for a total of 1 minute at which time if the RSUs mode has not changed successfully we can pretty confidently say something has gone wrong on the RSU.

paulbourelly999: I am open to making this longer, say 2 minutes of polling before an error if that addresses your concern

jwillmartin: I'm cool with it being as-is if it works now. If we see an issue with other RSUs, then we can change it to a greater value.

jwillmartin (approved):

PR #756, opened 2025-08-18
Free message frame pointer

paulbourelly999:
# PR Details
## Description
Free message frame pointer to avoid memory leak

## Motivation and Context
avoid memory leak

## How Has This Been Tested?
Unit testing and Integration testing 

## Types of changes

## Checklist:

### DIFF
--- a/src/tmx/TmxApi/tmx/messages/TmxJ2735Codec.hpp
+++ b/src/tmx/TmxApi/tmx/messages/TmxJ2735Codec.hpp
@@ -361,8 +361,10 @@ public:
 				{
 					MessageFrameMessage *frame = TmxJ2735EncodedMessage<MessageFrameMessage>::decode_j2735_message<
 							codec::uper<MessageFrameMessage> >(theData);
-					if (frame)
+					if (frame) {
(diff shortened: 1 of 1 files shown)


src/tmx/TmxApi/tmx/messages/TmxJ2735Codec.hpp
```diff
@@ -361,8 +361,10 @@ class TmxJ2735EncodedMessage: public TmxJ2735EncodedMessageBase
 				{
 					MessageFrameMessage *frame = TmxJ2735EncodedMessage<MessageFrameMessage>::decode_j2735_message<
 							codec::uper<MessageFrameMessage> >(theData);
-					if (frame)
+					if (frame) {
 						_decoded.reset(new MsgType(frame->get_j2735_data()));
+					}
+					free(frame);
```
DokurOmkar, line 367: Do you think that message frame pointer needs to be freed in other places as wekk (ex: encode_j2735_message())?

kjrush (changes requested):
image
Actually, looking at it further, this sonar comment does seem to be accurate? The memory here does seem to come from `new` so we should `delete` it instead of `free`

PR #761, opened 2025-09-11
Fix SpatPlugin mem leak

paulbourelly999:
# PR Details
## Description
Pull Request to fix SPAT Plugin Memory leak.

## How Has This Been Tested?
Tested with virtual signal controller sending both
- 2007 ICD/TSCBM and j2735 2016 SPAT

## Types of changes

## Checklist:

### DIFF
--- a/src/tmx/TmxApi/tmx/j2735_messages/J2735MessageFactory.hpp
+++ b/src/tmx/TmxApi/tmx/j2735_messages/J2735MessageFactory.hpp
@@ -221,7 +221,12 @@ public:
 		}
 	}
 
-	~J2735MessageFactory()	{ }
+	~J2735MessageFactory()	{ 
+		for (auto &pair : byInt) {
+			if (pair.second)
+				delete pair.second;
+		}
+	}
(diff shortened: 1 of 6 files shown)


src/tmx/TmxApi/tmx/j2735_messages/J2735MessageFactory.hpp
```diff
@@ -221,7 +221,12 @@ class J2735MessageFactory
 		}
 	}
 
-	~J2735MessageFactory()	{ }
+	~J2735MessageFactory()	{
```
tsprinkmeier, line 224: should this be `virtual`? is the factory ever going to be sub-classed?

paulbourelly999: I don't think so. I can't imagine a case in which we would sub-class this factory and not just update it but we could make it virtual just for extensibilities sake so that if external developers want to modify this for their own use they can. I think that is a good idea

PR #762, opened 2025-09-16
Convert String into JSON for telematic JSON payload

dan-du-car:
# PR Details
## Description
Telematic plugin sends the SensorDetectedObject payload as a string of JSON content. Telematic tool treat this content as a whole string and do not read each individual field value nor save them correctly in the database. This fix is to convert the payload content into JSON type, so that telematic tool can properly parse the JSON.

## Related GitHub Issue
NA

## Related Jira Key
NA

## How Has This Been Tested?
Local integration testing

## Types of changes

## Checklist:


src/v2i-hub/TelematicBridgePlugin/test/TestTelematicMsgWorker.cpp
paulbourelly999, line 1: Can we add one unit test to run the routableMessageToJsonValue() function with a created SensorDetectedObject(). This was if someone would reintroduce this bug or a similiar but we would catch it in until tests. Currently this is not caught by our existing unit tests

PR #771, opened 2025-10-01
Temporary fix for SPAT Plugin memory leak.

paulbourelly999:
# PR Details
## Description
This is a temporary fix for a large memory leak in the SPAT Plugin. This is largely due to the complexity of memory management with the V2X provided wrapper classes for J2735 ASN1 Compiler generate C structs TmxJ2735Message and TmxJ2735EncodeMessage and the ASN1 Compiler generated structs incompatibility with smart pointers.

The V2X provided wrapper classes for J2735 ASN1 Compiler generated C Structs `TmxJ2735Message` and `TmxJ2735EncodedMessage` both store a shared pointer to the underlying C struct. Depending on how they are created they will clean up none or part of this object. For instance :
**From V2X Hub Plugin programing guide**
`TmxJ2735Message(DataType &msg)`
A constructor from an existing J2735 message reference.  In this case, it is assumed that memory management is done outside this class, so destroying the object has no impact on the underlying structure.
`TmxJ2735Message(DataType *data = 0)`
This constructor builds a J2735 message from the given pointer to an already created structure.  This form assumes control over the management of the pointer, i.e. it will be freed when the object is destructed.  If the pointer is null, which is the default case, the message is considered empty.
`TmxJ2735Message(const TmxJ2735Message <DataType> & msg)`
A basic copy constructor that directly uses the same underlying J2735 structure pointer, effectively increasing the reference count by one.  Therefore, the pointer will only be freed when the last reference to it was destroyed.  Therefore, the original object can go out of scope without destroying the underlying structure used in this copy. Because the pointers are shared, changes to data inside will affect all references.  

With these varying constructors it is easy to create these wrapper types without correctly freeing their underlying C Structs

Additionally using smart pointers with the ASN1 Compiler generated C structs causes difficult with memory management. This is because they do not have destructors and instead rely on provide macros `ASN_STRUCT_FREE` to recursively step throw members and free all allocated memory. Using smart pointers will cause segmentation faults when the smart pointer destructor is called if the underlying pointer was already freed the provided macros and will leak memory if the macros is not called since the member of the C struct are not all recursively freed.

## Related Jira Key
VH-1449

## How Has This 
[Screencast from 10-01-2025 03:11:52 PM.webm](https://github.com/user-attachments/assets/8a6bbdf1-a7f7-4c9c-9252-4315ac77c4a0)
Been Tested?
Local Integration Testing with Econolite Virtual Signal Controller. Video Included below.

## Types of changes

## Checklist:

### DIFF
--- a/src/v2i-hub/SpatPlugin/src/SignalControllerConnection.cpp
+++ b/src/v2i-hub/SpatPlugin/src/SignalControllerConnection.cpp
@@ -37,7 +37,7 @@ namespace SpatPlugin {
         return status;
     };
 
-    void SignalControllerConnection::receiveBinarySPAT(const std::shared_ptr<SPAT> &spat, uint64_t timeMs ) const {
+    void SignalControllerConnection::receiveBinarySPAT(SPAT *spat, uint64_t timeMs ) const {
         FILE_LOG(tmx::utils::logDEBUG) << "Receiving binary SPAT ..." << std::endl;
         char buf[SPAT_BINARY_BUFFER_SIZE];
         auto numBytes = spatPacketReceiver->TimedReceive(buf, SPAT_BINARY_BUFFER_SIZE, UDP_SERVER_TIMEOUT_MS);
@@ -47,10 +47,10 @@ namespace SpatPlugin {
             Ntcip1202 ntcip1202;
             ntcip1202.setSignalGroupMappingList(this->signalGroupMapping);
             ntcip1202.copyBytesIntoNtcip1202(buf, numBytes);
-            ntcip1202.ToJ2735SPAT(spat.get(),timeMs, intersectionName, intersectionId);
+            ntcip1202.ToJ2735SPAT(spat,timeMs, intersectionName, intersectionId);
             if (tmx::utils::FILELog::ReportingLevel() >= tmx::utils::logDEBUG)
             {
-                xer_fprint(stdout, &asn_DEF_SPAT, spat.get());
+                xer_fprint(stdout, &asn_DEF_SPAT, spat);
             }
         }
         else {
--- a/src/v2i-hub/SpatPlugin/src/SignalControllerConnection.cpp
+++ b/src/v2i-hub/SpatPlugin/src/SignalControllerConnection.cpp
@@ -37,7 +37,7 @@ namespace SpatPlugin {
         return status;
     };
 
-    void SignalControllerConnection::receiveBinarySPAT(SPAT *spat, uint64_t timeMs ) const {
+    void SignalControllerConnection::receiveBinarySPAT(SPAT * const spat, uint64_t timeMs ) const {
         FILE_LOG(tmx::utils::logDEBUG) << "Receiving binary SPAT ..." << std::endl;
         char buf[SPAT_BINARY_BUFFER_SIZE];
         auto numBytes = spatPacketReceiver->TimedReceive(buf, SPAT_BINARY_BUFFER_SIZE, UDP_SERVER_TIMEOUT_MS);
--- a/src/v2i-hub/SpatPlugin/src/SignalControllerConnection.h
+++ b/src/v2i-hub/SpatPlugin/src/SignalControllerConnection.h
@@ -94,7 +94,7 @@ namespace SpatPlugin {
              * @param spat an empty SPaT pointer to which the SPAT data will be written.
              * @param timeMs current time in ms from epoch to use for message timestamp.
              */
-            void receiveBinarySPAT( SPAT *spat, uint64_t timeMs) const;
+            void receiveBinarySPAT( SPAT * const spat, uint64_t timeMs) const;
             /**
              * @brief Method to receive SPaT data in UPER Hex format from TSC.
              * @param spatEncoded_ptr Empty SpatEncodedMessage to which the UPER encoded SPaT data will be written.
--- a/src/v2i-hub/SpatPlugin/src/SignalControllerConnection.h
+++ b/src/v2i-hub/SpatPlugin/src/SignalControllerConnection.h
@@ -94,7 +94,7 @@ namespace SpatPlugin {
              * @param spat an empty SPaT pointer to which the SPAT data will be written.
              * @param timeMs current time in ms from epoch to use for message timestamp.
              */
-            void receiveBinarySPAT(const std::shared_ptr<SPAT> &spat, uint64_t timeMs) const;
+            void receiveBinarySPAT( SPAT *spat, uint64_t timeMs) const;
             /**
              * @brief Method to receive SPaT data in UPER Hex format from TSC.
              * @param spatEncoded_ptr Empty SpatEncodedMessage to which the UPER encoded SPaT data will be written.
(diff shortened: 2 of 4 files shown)


src/v2i-hub/SpatPlugin/src/SpatPlugin.cpp
```diff
@@ -116,15 +116,23 @@ namespace SpatPlugin {
...
+					
+					tmx::messages::SpatMessage _spatMessage(*spat_ptr);
+					MessageFrameMessage frame(_spatMessage.get_j2735_data());
+					spatEncoded_ptr->set_data(TmxJ2735EncodedMessage<SPAT>::encode_j2735_message<codec::uper<MessageFrameMessage>>(frame));
 					spatEncoded_ptr->addDsrcMetadata(tmx::messages::api::msgPSID::signalPhaseAndTimingMessage_PSID);
 					auto rMsg = dynamic_cast<routeable_message*>(spatEncoded_ptr.get());
 					BroadcastMessage(*rMsg);
+					// Recursively free SPAT struct 
+					ASN_STRUCT_FREE(asn_DEF_SPAT, spat_ptr);
+					// TODO fix MessageFrameMessage destructor to properly free internal SPAT pointer
```
dan-du-car, line 132: Is this TODO still needed?

src/v2i-hub/SpatPlugin/test/TestSignalControllerConnection.cpp
```diff
@@ -335,11 +335,15 @@ namespace SpatPlugin {
...
+        // Free the allocated memory for the SPAT structure
+        ASN_STRUCT_FREE(asn_DEF_SPAT, spat);
     }
     TEST_F(TestSignalControllerConnection, receiveBinarySPATException) {
         EXPECT_CALL(*mockUdpServer, TimedReceive(_, 1000, 1000)).WillOnce(testing::DoAll( Return(0)));
-        auto spat = std::make_shared<SPAT>();
+        auto spat = (SPAT*)calloc(1, sizeof(SPAT));
 		EXPECT_THROW(signalControllerConnection->receiveBinarySPAT(spat, 1721238398773), tmx::utils::UdpServerRuntimeError);
+        // Free the allocated memory for the SPAT structure
+        ASN_STRUCT_FREE(asn_DEF_SPAT, spat);
```
dan-du-car, line 346: free(frame.get_j2735_data().get()); here as well?

src/v2i-hub/SpatPlugin/src/SpatPlugin.cpp
paulbourelly999: Yes this is the more robust fix we plan to implement on core.

src/v2i-hub/SpatPlugin/test/TestSignalControllerConnection.cpp
paulbourelly999: No we did not create a MessageFrameMessage here

src/v2i-hub/SpatPlugin/src/SignalControllerConnection.cpp
```diff
@@ -37,7 +37,7 @@ namespace SpatPlugin {
         return status;
     };
 
-    void SignalControllerConnection::receiveBinarySPAT(const std::shared_ptr<SPAT> &spat, uint64_t timeMs ) const {
+    void SignalControllerConnection::receiveBinarySPAT(SPAT *spat, uint64_t timeMs ) const {
```
kjrush: I see the prior ptr was const, should this be `const SPAT *...`?

src/v2i-hub/SpatPlugin/src/SignalControllerConnection.h
```diff
@@ -94,7 +94,7 @@ namespace SpatPlugin {
              * @param spat an empty SPaT pointer to which the SPAT data will be written.
              * @param timeMs current time in ms from epoch to use for message timestamp.
              */
-            void receiveBinarySPAT(const std::shared_ptr<SPAT> &spat, uint64_t timeMs) const;
+            void receiveBinarySPAT( SPAT *spat, uint64_t timeMs) const;
```
kjrush: Same as above RE: const

src/v2i-hub/SpatPlugin/src/SpatPlugin.cpp
```diff
@@ -116,15 +116,23 @@ namespace SpatPlugin {
...
 					PLOG(logDEBUG) << "Starting BINARY SPaT Receiver ...";
-					scConnection->receiveBinarySPAT(spat_ptr, PluginClientClockAware::getClock()->nowInMilliseconds());
-					tmx::messages::SpatMessage _spatMessage(spat_ptr);
 					auto spatEncoded_ptr = std::make_shared<tmx::messages::SpatEncodedMessage>();
-					spatEncoded_ptr->initialize(_spatMessage,"", 0U, IvpMsgFlags_RouteDSRC);
+					scConnection->receiveBinarySPAT(spat_ptr, PluginClientClockAware::getClock()->nowInMilliseconds());
+					
+					tmx::messages::SpatMessage _spatMessage(*spat_ptr);
+					MessageFrameMessage frame(_spatMessage.get_j2735_data());
+					spatEncoded_ptr->set_data(TmxJ2735EncodedMessage<SPAT>::encode_j2735_message<codec::uper<MessageFrameMessage>>(frame));
```
dan-du-car, line 126: What does that **V2xHub plugin programming guide** tell us about this `TmxJ2735EncodedMessage`? Which `TmxJ2735Message `or its related `TmxJ2735EncodedMessage `message type being used in this case?

src/v2i-hub/SpatPlugin/src/SpatPlugin.cpp
dan-du-car: Do we have a follow on story or github issue created for this plan?

src/v2i-hub/SpatPlugin/src/SignalControllerConnection.cpp
paulbourelly999: Before it was a const reference to a shared pointer. This means that the method did not increase the reference count for the smart pointer. This is done when passing a smart pointer to a function which should non outlive the calling function. Essentially this is sharing a smart pointer without passing ownership of the smart pointer to a function.  

This pointer is modified by this function so I don't think I can  make it const.

src/v2i-hub/SpatPlugin/src/SignalControllerConnection.h
paulbourelly999: see response above

dan-du-car (approved):

src/v2i-hub/SpatPlugin/src/SpatPlugin.cpp
paulbourelly999: the `spatEncoded_ptr` is a `TmxJ2735EncodedMessage<SPAT>`,
` _spatMessage` and `frame` are `TmxJ2735Message<SPAT>`  and `TmxJ2735Message<MessageFrame>` respectively

paulbourelly999: If you want to read more about these message types the programing guide is included on github and on confluence

src/v2i-hub/SpatPlugin/src/SignalControllerConnection.cpp
paulbourelly999: Corrected to this syntax ` SPAT *  const spat` which gives you a const pointer where the pointer can not be modified to point to a difference address but the value can be modified (https://www.geeksforgeeks.org/c/difference-between-const-int-const-int-const-and-int-const/).

Issue #749, opened 2025-07-31
ImmediateForward plugin unable to set RSU operating mode on RSU


charlewong:
### **Employee Status** (Select one)

No

### Summary

I have installed V2X-Hub 7.9.0 on an Ubuntu 22.04.6 machine from a git checkout and have configured the Immediate Forward Plugin to communicate with  my RSU via NTCIP 1218. I have set the snmp credentials for my RSU.  The V2X-Hub logs show that the ImmediateForward plugin is unable to set the RSU operating mode over snmp.

### Version

7.8.1 (Current)

### Expected Behavior

The plugin should be able to set the RSU operating mode on the RSU with SNMP

### Actual Behavior

The Event Log tab displays the following warning:
```
ImmediateForward encountered unhandled exception: Failed to set RSU to operational mode backtrace: (Hint: Use addr2line -C -e 0x#######) to find line number) /var/www/plugins/ImmediateForwardPlugin/bin/ImmediateForwardPlugin(+0x1a300) [0x400001a300] /var/www/plugins/ImmediateForwardPlugin/bin/ImmediateForwardPlugin(+0x1249c) [0x400001249c] /var/www/plugins/ImmediateForwardPlugin/bin/ImmediateForwardPlugin(+0x242b8) [0x40000242b8] /var/www/plugins/ImmediateForwardPlugin/bin/ImmediateForwardPlugin(+0x33dbc) [0x4000033dbc] /usr/local/lib/libtmxapi.so(ivp_onMessageReceived+0x138) [0x4001e46018] /usr/local/lib/libtmxapi.so(+0x61f4) [0x4001e461f4] /usr/local/lib/libtmxapi.so(ivp_receive+0x10) [0x4001e46354] /lib/aarch64-linux-gnu/libc.so.6(+0x7d5b8) [0x400239d5b8] /lib/aarch64-linux-gnu/libc.so.6(+0xe5edc) [0x4002405edc] diagnostic info: Throw location unknown (consider using BOOST_THROW_EXCEPTION) Dynamic exception type: tmx::TmxException std::exception::what: Failed to set RSU to operational mode
```

The v2xhub docker container logs:
```
[...]

PR #774, opened 2025-10-06
Remove unnecessary manual configuration steps

paulbourelly999:
# PR Details
## Description
This PR attempts to get V2X Hub closer to a state where V2X Hub is deployable with only a DB configuration state (localhost.sql file) and a docker environment variable state (.env file). It attempts to do this by:

- Replace secret files with docker env secrets
- Use deployment docker compose for dev container
- add docker env to create admin user
- Fix add_user and update_user queries
- Add script to generate certificates in case non exist 

## Related GitHub Issue
Also fixes issues described in https://github.com/usdot-fhwa-OPS/V2X-Hub/issues/749 at the bottom. By making the Dev Container setup use the deployment docker compose file, it has access to any folders/volumes created under the configuration/ directory so having duplicate volumes for devcontainer and for regular deployment is no longer necessary

## Related Jira Key
NA

## Motivation and Context
Improve deployment by reducing manual steps necessary for deployment and limiting all configuration to DB state and docker environment state

## How Has This Been Tested?
Local integration testing with Dev Container. 

**Steps to test regular deployment**

1. Navigate to `configuration/`
2. Clear any certificates in `configuration/ssl/` directory
3. Run V2X via `configuration/initialization.sh` script
4. Ensure V2X Hub UI deploys and is accessible

**Steps to test develop container deployment**

1. Clear any certificates in `configuration/ssl/` directory
2. Open V2X Hub repository using VSCode
3. Launch Dev Container using CTRL + P and searching "Dev Containers:Rebuild Container"
4. Build code using CTRL + SHIFT + B
5. Run V2X Hub by navigating to the "Terminal" dropdown in VSCode, selecting the "Run Task..." options and selecting the "run" task
6. Navigate to https://127.0.0.1:19760 to accept certificate warning 
7. Navigate to https://127.0.0.1 to access V2X Hub UI
8. Ensure V2X Hub UI is functional

## Types of changes

## Checklist:


.devcontainer/docker-compose-vscode.yml
paulbourelly999, line 1: Updated the devcontainer to merge deployment docker compose with dev container docker compose. This means I only need to define modifications or additions in the devcontainer docker compose. Removing everything else that is duplicate

.devcontainer/devcontainer.json
```diff
@@ -6,7 +6,7 @@
 	"name": "V2X Hub Docker Compose Dev",
 
 	// Update the 'dockerComposeFile' list if you have more compose files or use different names.
-	"dockerComposeFile": "docker-compose-vscode.yml",
+	"dockerComposeFile": ["../configuration/docker-compose.yml","docker-compose-vscode.yml",],
```
paulbourelly999, line 9: See docker compose merge functionality documentation (https://docs.docker.com/compose/how-tos/multiple-compose-files/merge/)

.devcontainer/docker-compose-vscode.yml
```diff
@@ -4,120 +4,15 @@ services:
...
-    mysql_password:
-        file: ../secrets/mysql_password.txt
-    mysql_root_password:
-        file: ../secrets/mysql_root_password.txt
-
-volumes:
-  mysql-datavolume:
-  kafka-datavolume:
-  zookeeper-datavolume:
+      - ../.devcontainer/mysql_add_dev_user.sql:/docker-entrypoint-initdb.d/mysql_add_dev_user.sql
```
paulbourelly999, line 18: Path is not relative to deployment docker-compose file since it is the first in the heirarchy

configuration/docker-compose.yml
```diff
@@ -6,11 +6,10 @@ services:
       - MYSQL_DATABASE=IVP
       - MYSQL_USER=IVP
       - MYSQL_PASSWORD_FILE=/run/secrets/mysql_password
-      - MYSQL_ROOT_PASSWORD_FILE=/run/secrets/mysql_root_password
+      - MYSQL_RANDOM_ROOT_PASSWORD=yes
     network_mode: host
     secrets:
       - mysql_password
-      - mysql_root_password
```
paulbourelly999, line 13: Removed mysql_root_password. V2X Hub uses an admin privileged IVP mysql user. Removing root password definition means that getting credentials to IVP user, will not also be root user credentials. Root user credentials are now randomly initialized and can be found only in mysql docker logs.

configuration/initialization.sh
```diff
@@ -3,8 +3,6 @@ set -e
 
 # Initialize Docker Environment for V2X Hub
 ./initialize_docker_environment.sh
-# Initialize secrets
-./initialize_secrets.sh
```
paulbourelly999, line 7: Since secret files are no longer necessary, initialize secrets is also no longer necessary

configuration/initialize_docker_environment.sh
```diff
@@ -72,7 +72,27 @@ if [[ $SIMULATION_MODE == "TRUE" ]]; then
     read -r -p "Enter Sensor Configuration File Path (or press Enter to use default as $SENSOR_JSON_FILE_PATH_DEFAULT): " SENSOR_JSON_FILE_PATH
     SENSOR_JSON_FILE_PATH=${SENSOR_JSON_FILE_PATH:-$SENSOR_JSON_FILE_PATH_DEFAULT}
 fi
-
+# MySQL (IVP) User Password
+read -r -p "MYSQL PASSWORD (TRUE/FALSE, or press Enter to use default as $MYSQL_PASSWORD_DEFAULT): " MYSQL_PASSWORD
```
paulbourelly999: Adding prompts to read mysql password, v2x username and password

configuration/README.md
```diff
@@ -50,6 +50,8 @@ To support execution in a simulated environment, V2X-Hub is in the process of in
 * **SIM_V2X_PORT** – Environment variable for storing port for sending v2x messages to CDASim
 * **V2XHUB_IP** – Environment variable for storing IP address of V2X Hub.
 * **INFRASTRUCTURE_ID** – Environment variable for storing infrastructure id of V2X Hub.
+* **V2XHUB_USER** – V2X Hub Administrator Username to create on startup
+* **V2XHUB_USER** – V2X Hub Administrator Password to create on startup
```
willjohnsonk: Should this be "V2XHUB_USERNAME" and "V2XHUB_PASSWORD" based on the actual env variables?

Also nit while in this file- should we be using our internal, default user/pass as an example in the README on line 24/26?

willjohnsonk: Also, should the README be updated to reflect changes in the DB password? It looks like it is being set to a random value initially.

configuration/initialization.sh
willjohnsonk: If we are removing secrets I assume documentation in several places may need updates to reflect the change. Just a reminder

container/service.sh
```diff
@@ -1,14 +1,15 @@
...
 # Add stol apt repository libraries to path for tmxcore
 export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/opt/carma/lib/
 for plugin in /usr/local/plugins/*.zip; do
     echo "Installing plugin $plugin"
     tmxctl --plugin-install "$plugin"
 done
-
+cd container/
+# Generate self-signed certificates if they do not already exist
+./generate_certificates.sh
```
willjohnsonk, line 12: I assume these certs are for the browser? Does this solve the issue where we previously were launching chromium with the no certs argument? Again, worth double checking if this needs documentation updates elsewhere

willjohnsonk (changes requested):

configuration/README.md
paulbourelly999: The user will still be prompted for a V2X Hub username and password which will be written as docker environment variables described here so I think 24/26 is still accurate. Additionally the MYSQL_PASSWORD still also exists and the user is prompted for it but the user is no longer prompted for a MYSQL_ROOT_PASSWORD which is now randomly generated. Ideally we should not be using this root user at all. I will update line 18 to reflect that we no longer prompt users for MYSQL_ROOT_PASSWORD.

paulbourelly999: I think I addressed all the concerns here. Please let me know if I missed anything.

configuration/initialization.sh
paulbourelly999: I don't think the documentation calls out these secrets at all. Also we are not removing them. They were previously initialized via txt file. We are now using docker environment variables to initialize them. This makes deployment easier since fewer separate configuration files are necessary.

container/service.sh
paulbourelly999: Yeah so that issue has already been solved using mkcerts and documented in the `configuration/README.md` at the bottom with a note. This does not solve that issue. This will generate certificates that still have  that same self-signed issue, but only if none are already present. This is something that helps V2X Hub require less manual steps to deploy.

In other words if users want to go through the process of generated trusted certificates they can deploy V2X Hub with the initialization script which will run mkcerts to generate trusted certificates.

If they just want to deploy V2X Hub and already have certificates or are ok with self-signed certificates they can simply run the container and it will check to see if certificates exists and generate self-signed ones that need to be explicitly trusted by the browser

configuration/initialization.sh
willjohnsonk: I specifically remember a setup guide that mentions manually creating the secrets text file as a step. May be related to some of the material Will made for CAVe or on Confluence? I will look around, just don't want outdated instructions being carried forward

configuration/README.md
willjohnsonk: > The user will still be prompted for a V2X Hub username and password which will be written as docker environment variables described here so I think 24/26 is still accurate. 

My comment here was more on the use of "V2xHub#321" as the example password when that is the one we use so often in the lab. I know security isn't that much of a concern but it stood out

Any comments on the naming of variables in line 52/53?

configuration/initialization.sh
paulbourelly999: I think that is completely out of date since the initialization script would create those files for you before this change. We can delete those instructions completely.

configuration/README.md
paulbourelly999: Good call, I can address that.

willjohnsonk (approved):
💯

PR #864, opened 2026-04-08
Update/mysql hardened image

jwillmartin:
# PR Details
## Description

Made updates to use the MySQL 9.6 Docker Hardened Image. This update required changes to the docker-compose.yml files, initialization scripts, and a new entrypoint wrapper to handle the behavior change between images.

## Motivation and Context

The previous MySQL images contained many known vulnerabilities and was a few versions behind the latest release. This update moves the image to the latest release, plus adds additional security by using a hardened image.

## How Has This Been Tested?

Tested using the initialization script and using the dev container. Both methods correctly came up. 
Verified the MAP and Immediate Forward plugins functioned without deviation.

## Types of changes

## Checklist:


configuration/mysql/entrypoint-wrapper.sh
```diff
@@ -0,0 +1,47 @@
...
+    if [ -n "${MYSQL_DATABASE:-}" ]; then
+        echo "init: Creating database '${MYSQL_DATABASE}'"
+        mysql -h 127.0.0.1 -u root -p"${MYSQL_ROOT_PASSWORD}" \
+            <<< "CREATE DATABASE IF NOT EXISTS \`${MYSQL_DATABASE}\`;"
+    fi
+
+    # Create user and grant access
+    if [ -n "${MYSQL_USER:-}" ] && [ -n "${MYSQL_PASSWORD:-}" ]; then
+        echo "init: Creating user '${MYSQL_USER}'"
+        mysql -h 127.0.0.1 -u root -p"${MYSQL_ROOT_PASSWORD}" <<-EOSQL
```
paulbourelly999, line 29: Apparently we should not even need this script. We should still be able to use the .sql files the same way, except we may need to create the initial database using and environment variable like described below. Am I missing something here? Why do we need this new script

Image

PR #867, opened 2026-04-13
Remove unnessary dependencies

paulbourelly999:
# PR Details
## Description
**Before changes:**
image
**After changes:**
image

Roughly 127 mb reduction in compressed image size ~25 %

This was accomplished by 
- removing build depedencies from deployment image like build-essentials
- replacing libboost-all-dev with specific libboost packages
- update Dockerfile `COPY` commands to only copy required compiled artifacts

## Related Jira Key
[VH-1415
](https://usdot-carma.atlassian.net/browse/VH-1514)

## Motivation and Context
Reducing production image size improves performance, security and usability by taking up less storage, decreasing the attack surface, and allowing quicker build or download of application

## How Has This Been Tested?
CI Unit tests and Integration Tests

## Types of changes

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,14 +38,19 @@ WORKDIR /home/V2X-Hub/container/
 RUN ./database.sh && ./library.sh && ldconfig
 # Built Plugins
 COPY --from=dependencies --chown=plugin:adm --chmod=544 /usr/local/plugins/ /usr/local/plugins/
-# Built Libraries for V2X Hub (tmx services) and ext/ (snmp, etc)
-COPY --from=dependencies  --chown=plugin:adm --chmod=544 /usr/local/lib/libpdclient.a /usr/local/lib/
-COPY --from=dependencies  --chown=plugin:adm --chmod=544 /usr/local/lib/libqserverPedestrian.a /usr/local/lib/
+# Built Libraries for V2X Hub (tmx services)
 COPY --from=dependencies  --chown=plugin:adm --chmod=544 /usr/local/lib/libtmxapi.* /usr/local/lib/
 COPY --from=dependencies  --chown=plugin:adm --chmod=544 /usr/local/lib/libtmxctlStatic.a /usr/local/lib/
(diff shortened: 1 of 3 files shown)


kelleydi:
@paulbourelly999 : Should you update the documentation in `/src/v2i-hub/docs/Programming_Guide.md`? Maybe add a note that states 


> ⚠️ Dependency Management and Boost Libraries
> 
> The V2X Hub runtime environment does not include the full `libboost-all-dev` package. Instead, a minimal subset of Boost libraries is installed based on current system and plugin requirements.
> 
> - Do not assume all Boost libraries are available in the runtime environment.
> - When adding or modifying a plugin:
>   - Explicitly identify required Boost components
>   - Ensure corresponding runtime libraries are included in the deployment image
> - If a plugin introduces new Boost dependencies, the Docker build and deployment scripts must be updated accordingly.

paulbourelly999:
> @paulbourelly999 : Should you update the documentation in `/src/v2i-hub/docs/Programming_Guide.md`? Maybe add a note that states
> 
> > ⚠️ Dependency Management and Boost Libraries
> > The V2X Hub runtime environment does not include the full `libboost-all-dev` package. Instead, a minimal subset of Boost libraries is installed based on current system and plugin requirements.
> > 
> > * Do not assume all Boost libraries are available in the runtime environment.
> > * When adding or modifying a plugin:
> >   
> >   * Explicitly identify required Boost components
> >   * Ensure corresponding runtime libraries are included in the deployment image
> > * If a plugin introduces new Boost dependencies, the Docker build and deployment scripts must be updated accordingly.

I think instead of making this specific to boost dependencies, we can come up with guidance on how to include any dependency. I do want to postpone this documentation until we come up with a specific process for including depedencies. Specifically I am looking to come up with a process long term that :

- allows for each package to document exactly which external depedencies it has
- each package should independently install dependencies especially optional components like plugins

Once we have this, generating documentation on how to include new depedencies will present more value

kelleydi (approved):
Looks solid.

PR #873, opened 2026-04-14
Fix/snmp client

jwillmartin:
# PR Details
## Description

These changes correct the SNMP Client implementation for versions 1 and 2. Additionally, there is a fix to the default timeout, security model for v1 and v2, and existing logging. 

## Motivation and Context

Issues were encountered when implementing SNMP -v1 in a plugin that uses the SNMPClient:
- Client with default timeout was timing out in 0.1 ms
- Logging octet string values came up empty
- Community string was not correctly declared
- Incorrect security model was used for v1 and v2

## How Has This Been Tested?

Built and tested locally with upcoming priority plugin. 
Updated log outputs after correct SNMP v1 implementation:

SNMPClient.cpp (270) - DEBUG1 : Attempting to SET octet string: 0169dea5de69dea5e10200000000000000000001000000000000000000010000000000000000000100000000000000000001000000000000000000010000000000000000000100000000000000000001000000000000000000010000000000000000000100000000000000000000 (110 bytes) for OID: 1.3.6.1.4.1.1206.4.2.11.4.1.0
SNMPClient.cpp (353) - DEBUG  : Success in SET for OID: 1.3.6.1.4.1.1206.4.2.11.4.1.0 Value: 0169dea5de69dea5e10200000000000000000001000000000000000000010000000000000000000100000000000000000001000000000000000000010000000000000000000100000000000000000001000000000000000000010000000000000000000100000000000000000000 (110 bytes)

## Types of changes

## Checklist:

### DIFF
--- a/src/tmx/TmxUtils/src/SNMPClient.cpp
+++ b/src/tmx/TmxUtils/src/SNMPClient.cpp
@@ -1,4 +1,5 @@
 #include "SNMPClient.h"
+#include <tmx/messages/byte_stream.hpp>
 
 namespace tmx::utils
 {
@@ -20,15 +21,21 @@ namespace tmx::utils
         snmp_sess_init(&session);
         session.peername = ip_port;
         session.version = snmp_version_; // SNMP_VERSION_3
-        session.securityName = (char *)snmp_user.c_str();
-        session.securityNameLen = snmp_user.length();
-        session.securityModel = SNMP_SEC_MODEL_USM;
-
         // Fallback behavior to setup a community for SNMP V1/V2
         if (snmp_version_ != SNMP_VERSION_3)
         {
-            session.community = (unsigned char *)community.c_str();
+            session.community = (unsigned char *)community_.c_str();
             session.community_len = community_.length();
+            if (snmp_version_ == SNMP_VERSION_1)
+                session.securityModel = SNMP_SEC_MODEL_SNMPv1;
+            else
+                session.securityModel = SNMP_SEC_MODEL_SNMPv2c;
+        }
+        else
+        {
+            session.securityName = (char *)snmp_user.c_str();
+            session.securityNameLen = snmp_user.length();
+            session.securityModel = SNMP_SEC_MODEL_USM;
         }
 
         // Set security level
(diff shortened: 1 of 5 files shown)


src/tmx/TmxUtils/src/SNMPClient.cpp
```diff
@@ -255,17 +273,16 @@ namespace tmx::utils
...
                 }
-                // Needs to be finalized to support octet string use
                 else if (val.type == snmp_response_obj::response_type::STRING)
                 {
-
-                    std::string val_string (val.val_string.begin(), val.val_string.end());
-                    PLOG(logDEBUG1) << "Attempting to SET value for " << input_oid << " to " << val_string;
-
-                    snmp_add_var(pdu, OID, OID_len, 's', val_string.c_str());
+                    PLOG(logDEBUG1) << "Attempting to SET octet string: "
```
paulbourelly999, line 288: Is this the only reason we have imported `tmx::byte_stream_encode`. What is the difference in output between the previous log statement an this one?

src/tmx/TmxUtils/src/SNMPClient.cpp
```diff
@@ -20,15 +30,21 @@ namespace tmx::utils
...
-
         // Fallback behavior to setup a community for SNMP V1/V2
         if (snmp_version_ != SNMP_VERSION_3)
         {
-            session.community = (unsigned char *)community.c_str();
+            session.community = reinterpret_cast<unsigned char *>(community_.data());
             session.community_len = community_.length();
+            if (snmp_version_ == SNMP_VERSION_1)
+                session.securityModel = SNMP_SEC_MODEL_SNMPv1;
+            else
```
paulbourelly999, line 42: The current code seems to default to version 2 if no version is specified. I do not think that is what we want to do.

src/tmx/TmxUtils/src/SNMPClient.cpp
```diff
@@ -20,15 +30,21 @@ namespace tmx::utils
...
-        session.securityNameLen = snmp_user.length();
-        session.securityModel = SNMP_SEC_MODEL_USM;
-
         // Fallback behavior to setup a community for SNMP V1/V2
         if (snmp_version_ != SNMP_VERSION_3)
         {
-            session.community = (unsigned char *)community.c_str();
+            session.community = reinterpret_cast<unsigned char *>(community_.data());
             session.community_len = community_.length();
+            if (snmp_version_ == SNMP_VERSION_1)
```
paulbourelly999, line 38: For if statement readability, lets always include {} curly braces

src/tmx/TmxUtils/src/SNMPClient.cpp
```diff
@@ -20,15 +30,21 @@ namespace tmx::utils
...
+                session.securityModel = SNMP_SEC_MODEL_SNMPv1;
+            else
+                session.securityModel = SNMP_SEC_MODEL_SNMPv2c;
```
paulbourelly999, line 44: For if statement readability lets always include curly braces

paulbourelly999 (changes requested):
Left some small comments

dan-du-car (dismissed):

src/tmx/TmxUtils/src/SNMPClient.cpp
jwillmartin: Yeah, this is why. It wasn't outputting anything before.

src/tmx/TmxUtils/src/SNMPClient.cpp
jwillmartin: It still defaults to version 3, if nothing is specified. The default is declared in the header file.
These were added to set the correct security model for v1 and v2.

src/tmx/TmxUtils/src/SNMPClient.cpp
jwillmartin: Added.

src/tmx/TmxUtils/src/SNMPClient.cpp
jwillmartin: Added.

src/tmx/TmxUtils/src/SNMPClient.cpp
paulbourelly999: That is strange because for the latest release these log statements where definitely outputting data. I will try to confirm

paulbourelly999 (approved):

PR #888, opened 2026-05-08
Feature/priority plugin

jwillmartin:
# PR Details
## Description

Introducing new Priority Plugin, with all necessary updates to include it in V2X Hub and some additional scripts and tests.

## Related Jira Key
[HELPDESK-196
](https://usdot-carma.atlassian.net/browse/HELPDESK-196)

## Motivation and Context

The priority plugin was developed due to stakeholder demand. Two notable requests have been incoming for a plugin to handle and process both SRMs and SSMs. This plugin addresses those requests.

## How Has This Been Tested?

Tested with an HIL setup and with a live deployment. 
HIL:
Connected to a Yunex M60 controller over a VPN and send mock SRMs, using the mock-srm.py script, to the plugin to enable requests to the controller. Controller updates were observed, using the controller's terminal-based front panel, to verify against the decoded contoller responses in the Docker logs. Docker logs were compared with V2X Hub's JSON logs of the encoded SSMs.

Live deployment:
Tested in the field with a Commsignia OBU, various Commsignia RSUs, and various Yunex Blade controllers. The Commsignia OBU auto-generated SRMs as it drove through this live corridor and the RSUs forwarded the received SRMs back to V2X Hub. All controller front panels were observed while testing and compared against the docker logs. Docker logs were again compared with V2X Hub JSON logs.

## Types of changes

## Checklist:

- [x] I have added any new packages to the sonar-scanner.properties file


src/tmx/TmxApi/tmx/TmxApiMessages.h
```diff
@@ -238,7 +238,7 @@ static CONSTEXPR const char *STATUS_STOPPED_STRING = "Stopped / Disconnected";
 enum msgPSID
 {
 	None_PSID = 0x00,
-	mapData_PSID = 0x8002,
```
dan-du-car, line 241: Why choose this `0xE0000017`?

src/v2i-hub/SpatPlugin/test/TestSignalControllerConnection.cpp
dan-du-car, line 60: Do we need IP twice here "127.0.0.1?

jwillmartin: @dan-du-car
This was from the previous SNMP Client fixes to be able to handle SNMP v1. I hadn't merged develop into this branch yet. Just merged, so this should be gone now.

src/tmx/TmxApi/tmx/TmxApiMessages.h
jwillmartin: I noticed it was wrong while updating the PSID for SSM and just fixed it. Reference here:
https://standards.ieee.org/products-programs/regauth/psid/public/
Look for "MAP distribution"

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
```diff
@@ -0,0 +1,683 @@
...
+            } catch (const boost::property_tree::json_parser_error &e) {
+                PLOG(logERROR) << "Failed to parse LaneStrategyMapping JSON: " << e.what();
+            } catch (const boost::property_tree::ptree_error &e) {
+                PLOG(logERROR) << "Invalid LaneStrategyMapping entry: " << e.what();
+            }
+        }
+
+        // Parse the TSC_Controllers JSON array and create an SNMP client per entry
+        _controllers.clear();
+        try {
```
paulbourelly999, line 124: Lets move the parsing of the JSON configuration into a separate unit testable function somewhere. This will not only allow you to unit test this part of the functionality, but will also reduce the complexity of this update configuration method

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
```diff
@@ -0,0 +1,683 @@
...
+
+        // Clear strategy mapping JSON array
+        {
+            std::lock_guard lock(_tableMutex);
+            _processor.ClearLaneStrategyMap();
+        }
+        // Parse the LaneStrategyMapping JSON array and populate the processor's lane-strategy map
+        std::string laneStrategyJson;
+        GetConfigValue<std::string>("LaneStrategyMapping", laneStrategyJson);
+        if (!laneStrategyJson.empty()) {
```
paulbourelly999: Lets move this parsing of the configuration JSON into a separate unit testable method. This will not only allow you to increase the test coverage for this functionality. It will also reduce the complexity of the updateConfigurations method here.

Many plugins have implemented this by either creating free functions (functions without an associated class) or  a configuration class with static member functions ( member functions that can be called without having to instantiate an object of the class) in a separate header and CPP file called something like SomethingConfiguration.cpp/.hpp

See https://github.com/usdot-fhwa-OPS/V2X-Hub/blob/develop/src/v2i-hub/ImmediateForwardPlugin/src/ImmediateForwardConfiguration.cpp for reference

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
```diff
@@ -0,0 +1,683 @@
...
+    {
+        PLOG(logINFO) << "Received Signal Request Message (SRM)";
+        tmx::messages::SrmEncodedMessage srmEnc(routeableMsg);
+        auto srmDecoded = srmEnc.decode_j2735_message();
+        auto srm = srmDecoded.get_j2735_data();
+        if (!srm) {
+            PLOG(logWARNING) << "SRM decode returned null, skipping.";
+            _skippedMessages++;
+            SetStatus(_keySkippedMessages, _skippedMessages);
+            return;
```
paulbourelly999: NIT: Not a huge fan of using empty return statements in void functions to exit function because it make the function much more complex and difficult to read. While there are probably still some situations where this is cleaner, I think it is much more readable to either invert conditions like this 
```cpp
if (srm) {
            // all my valid logic 
}
else {
           PLOG(logWARNING) << "SRM contains no signal request packages, skipping.";
            _skippedMessages++;
            SetStatus(_keySkippedMessages, _skippedMessages);
}
```

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
```diff
@@ -0,0 +1,683 @@
...
+
+        if (!srm->requests || srm->requests->list.count <= 0) {
+            PLOG(logWARNING) << "SRM contains no signal request packages, skipping.";
+            _skippedMessages++;
+            SetStatus(_keySkippedMessages, _skippedMessages);
+            return;
+        }
+
+        // Extract the vehicle ID from the requestor
+        std::vector<uint8_t> vehicleID;
```
paulbourelly999: Similar to the updateConfig method above, it feels like much of this can be moved into a separate unit testable function to simplify this method, It may even help with the empty return statements since you could throw an exception or return a value from the function and based on that exception or value not broadcast the message

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
```diff
@@ -0,0 +1,683 @@
...
+        // Check reservice timer per 4.2.3.1 (h)
+        uint8_t classIdx = (classType >= 1 && classType <= 10) ? (classType - 1) : 9;
+        auto reservicePeriod = _reserviceClassTime[classIdx];
+        auto lastCompleted = _processor.ReserviceLastCompleted(classType);
+        if (reservicePeriod > 0 && lastCompleted > 0 &&
+            (static_cast<uint32_t>(nowEpoch) - lastCompleted) < reservicePeriod) {
+            entry.statusInPRS = RequestStatus::reserviceError;
+            PLOG(logINFO) << "Reservice period not met for class " << static_cast<int>(classType)
+                          << ", setting reserviceError for slot " << (freeIt - table.begin());
+            return;
```
paulbourelly999: NIT: Empty return statement, See comment above

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
```diff
@@ -0,0 +1,683 @@
...
+                                  timeOfRequest, state);
+            }
+        }
+
+        for (uint8_t i = 0; i < _maxSsmBroadcastsPerStatus; i++) {
+            BuildSSM(state);
+        }
+    }
+
+    void PriorityPlugin::ProcessPrsPackage(const SignalRequestPackage &pkg, const std::vector<uint8_t> &vehicleID, uint8_t classType, uint8_t classLevel, uint8_t newSeq, long role, long currentMinuteOfYear, long currentMsInMinute, time_t nowEpoch)
```
paulbourelly999: This entire function can probably go in a different CPP/Header file and does not need to be a member function of the PriorityPlugin. This way you can unit test it

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.hpp
```diff
@@ -0,0 +1,271 @@
...
+#include <TmxMessageManager.h>
+#include <tmx/messages/TmxJ2735.hpp>
+#include <tmx/messages/TmxJ2735Codec.hpp>
+#include <tmx/j2735_messages/SignalRequestMessage.hpp>
+#include <tmx/j2735_messages/SignalStatusMessage.hpp>
+
+namespace PriorityPlugin {
+
+    // NTCIP 1211 OID for prgPriorityRequestAbsolute (1211 v02A-SE03f PRS-MIB1 5.1.2.8)
+    static const std::string NTCIP1211_PRIORITY_REQUEST_ABSOLUTE_OID = "1.3.6.1.4.1.1206.4.2.11.2.8.0";
```
paulbourelly999: For NTCIP 1218 we created header files that contain all the OIDs incase other plugins or functionality needed access to this ODIs, We put it under src/tmx/TmxUtils/rsu/NTCIP_1218_MIB.h. I think doing something similar here under src/tmx/TmxUtils/tsc/NTCIP_1211_MIB.h would be awesome

src/v2i-hub/PriorityPlugin/src/PriorityPluginMessaging.cpp
```diff
@@ -0,0 +1,545 @@
...
+using namespace tmx::messages;
+using namespace tmx::utils;
+
+namespace PriorityPlugin {
+
+    void PriorityPlugin::ServiceExchangeLoop()
+    {
+        PLOG(logINFO) << "PRS service exchange loop started, poll interval = " << _pollIntervalMs << "ms";
+
+        while (_running) {
```
paulbourelly999: NIT: Instead of having the function in charge of sleep and the while loop. It is easier to read when the plugin is in charge of this and the function just represents a single execution of a loop. To illustrate the different

```cpp
void PriorityPlugin::doSomething() {
      PriorityPlugin::ServiceExchangeLoop();
      // Now I want to do something after but since this while loop do not return control to the calling function
      // not only will my statement not execute but whatever called my function will not return control either and 
     // will hold the thread. 
      PriorityPlugin::doSomethingElse()
} 
```

**Better Alternative for readability**
```cpp
void PriorityPlugin::doSomething() {
       // 2. Create the thread with a lambda that contains a loop
      std::thread worker([&_running]() {
          while (_running) {
               PriorityPlugin::ServiceExchangeOneLoop()
                std::this_thread::sleep_for(std::chrono::milliseconds(500));
          }
      });
     
      // Now I am done and want to do something else
     // 4. Signal the thread to stop and wait for it to finish
        _running= false;
        if (worker.joinable()) {
            worker.join(); // Blocks on return of running thread
        }
      PriorityPlugin::doSomethingElse()
} 
```
In the alternative I can clearly tell from reading this code that I have some sort of loop here and I can move the `join` call to decide where I want to block execution until this function returns, for example if I want to report some sort of error that has occurred.

src/v2i-hub/PriorityPlugin/src/PriorityPluginMessaging.cpp
```diff
@@ -0,0 +1,545 @@
...
+                }
+                // If no active requests, use first available controller (idle polling)
+                if (!targetClient && !_controllers.empty()) {
+                    targetClient = _controllers.begin()->second.snmpClient;
+                    targetIntersectionID = _controllers.begin()->second.intersectionID;
+                }
+            }
+
+            if (!targetClient) {
+                std::this_thread::sleep_for(std::chrono::milliseconds(_pollIntervalMs));
```
paulbourelly999: Avoid calls to thread sleep directly. These will not work in simulation. Please us the carma clock sleep instead which will default to the same behavior in non simulation but also works for simulation

src/v2i-hub/PriorityPlugin/src/PriorityPluginMessaging.cpp
```diff
@@ -0,0 +1,545 @@
...
+
+            _serviceExchanges++;
+            if (_serviceExchanges % 10 == 0) {
+                SetStatus(_keyServiceExchanges, _serviceExchanges);
+            }
+
+            // Broadcast SSM reflecting current table state
+            BroadcastSSMFromTable();
+
+            std::this_thread::sleep_for(std::chrono::milliseconds(_pollIntervalMs));
```
paulbourelly999: Avoid direct calls to sleep and use carma_clock sleep instead. See https://github.com/usdot-fhwa-OPS/V2X-Hub/blob/0f36b7c544f3f09e93fa03c792906fc73180784a/examples/tmx-exampleapps/ExamplePlugin/src/ExamplePlugin.cpp#L153

src/v2i-hub/PriorityPlugin/src/PriorityPluginMessaging.cpp
```diff
@@ -0,0 +1,545 @@
...
+                byIntersection[entry.intersectionID].push_back(&entry);
+            }
+        }
+
+        if (byIntersection.empty()) return;
+
+        auto ssmPtr = std::make_shared<SignalStatusMessage_t>();
+        memset(ssmPtr.get(), 0, sizeof(SignalStatusMessage_t));
+
+        time_t nowEpoch = std::time(nullptr);
```
paulbourelly999: Please us carma clock instead of direct calls to time

src/v2i-hub/PriorityPlugin/src/PriorityPluginMessaging.cpp
```diff
@@ -0,0 +1,545 @@
...
+            }
+
+            asn_sequence_add(&ssmPtr->status.list, signalStatus);
+        }
+
+        try {
+            SsmEncodedMessage encodedSSM;
+            MessageFrameMessage frame(ssmPtr);
+            encodedSSM.set_data(TmxJ2735EncodedMessage<SignalStatusMessage>::encode_j2735_message<codec::uper<MessageFrameMessage>>(frame));
+            free(frame.get_j2735_data().get()); // NOSONAR: ASN.1 C struct allocated via calloc
```
paulbourelly999: Instead of calling this manually you can register calls with the destructor of your smart pointer.  Something like this
```cpp
 std::shared_ptr<SignalStatusMessage_t> ptr(new SignalStatusMessage_t(), [](SignalStatusMessage_t* p) {
        ASN_STRUCT_FREE(asn_DEF_SignalStatusMessage, ptr);
 });
```

There exist a helper method for this delete operation in called j2735_destroy, which you can hand the ptr and message descript to to make it look like this 
```cpp
 std::shared_ptr<SignalStatusMessage_t> ptr(new SignalStatusMessage_t(), [](SignalStatusMessage_t* p) {
        tmx::messages::j2735::j2735_destroy( p, asn_DEF_SignalStatusMessage)
 });
```
The function you are currently using also does not look correct
```CPP
/*
 * Free the structure including freeing the memory pointed to by ptr itself.
 */
#define ASN_STRUCT_FREE(asn_DEF, ptr) \
    (asn_DEF).op->free_struct(&(asn_DEF), (ptr), ASFM_FREE_EVERYTHING)

/*
 * Free the memory used by the members of the structure without freeing the
 * the structure pointer itself.
 * ZERO-OUT the structure to the safe clean state.
 * (Retaining the pointer may be useful in case the structure is allocated
 *  statically or arranged on the stack, yet its elements are dynamic.)
 */
#define ASN_STRUCT_RESET(asn_DEF, ptr) \
    (asn_DEF).op->free_struct(&(asn_DEF), (ptr), ASFM_FREE_UNDERLYING_AND_RESET)

/*
 * Free memory used by the members of the structure without freeing
 * the structure pointer itself.
 * (Retaining the pointer may be useful in case the structure is allocated
 *  statically or arranged on the stack, yet its elements are dynamic.)
 * AVOID using it in the application code;
 * Use a safer ASN_STRUCT_RESET() instead.
 */
#define ASN_STRUCT_FREE_CONTENTS_ONLY(asn_DEF, ptr) \
    (asn_DEF).op->free_struct(&(asn_DEF), (ptr), ASFM_FREE_UNDERLYING)
```
Since we want to free everything here we should be calling ASN_STRUCT_FREE to avoid any memory leaks

src/v2i-hub/PriorityPlugin/src/PriorityPluginMessaging.cpp
jwillmartin: Addressed.

src/v2i-hub/PriorityPlugin/src/PriorityPluginMessaging.cpp
jwillmartin: Addressed

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
jwillmartin: Addressed, moved to PriorityConfiguration.cpp and added unit tests

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
jwillmartin: Addressed, moved to PriorityConfiguration.cpp and added unit tests

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
jwillmartin: Removed this specific return. Looking at addressing others.

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.hpp
jwillmartin: Addressed, moved to src/tmx/TmxUtils/tsc/NTCIP_1211_MIB.h

src/v2i-hub/PriorityPlugin/src/PriorityPluginMessaging.cpp
jwillmartin: Addressed

src/tmx/TmxApi/test/J2735MessageTest.cpp
```diff
@@ -568,6 +556,72 @@ TEST (J2735MessageTest, EncodeSrm)
...
+	sigStatus->status = 2;
+	asn_sequence_add(&signalStatus->sigStatus.list, sigStatus);
+	asn_sequence_add(&message->status.list, signalStatus);
+
+	tmx::messages::SsmEncodedMessage ssmEncodeMessage;
+	auto _ssmMessage = new tmx::messages::SsmMessage(message);
+	tmx::messages::MessageFrameMessage frame_msg(_ssmMessage->get_j2735_data());
+	ssmEncodeMessage.set_data(TmxJ2735EncodedMessage<SignalStatusMessage>::encode_j2735_message<codec::uper<MessageFrameMessage>>(frame_msg));
+	free(message);
+	free(frame_msg.get_j2735_data().get());
```
paulbourelly999: I think to properly free this data we should be using the ASN1_C provided macros. A good way to confirm whether you have properly freed the memory without memory leak is to compile V2X Hub with Debug flags, then run just this unit test using --gtest_filter=<test-case-name>. If this results in a ASan (address sanitizer) error its likely you have not correctly allocated or deallocated the memory

paulbourelly999: For reference this is the macros I was referring to:
```cpp
ASN_STRUCT_FREE((*asnType), ptr);
```

src/tmx/TmxApi/test/J2735MessageTest.cpp
```diff
@@ -568,6 +556,72 @@ TEST (J2735MessageTest, EncodeSrm)
...
+
+	// Test encoding
+	std::string expectedSsmEncHex = "001e1b6205cedea802000897b40b9004303028040a000437aa345989c408";
+	ASSERT_EQ(expectedSsmEncHex, ssmEncodeMessage.get_payload_str());
```
paulbourelly999: Use `EXPECT_EQ` instead of `ASSERT_EQ` normally. `EXPECT_EQ` generates non-fatal failures while `ASSERT_EQ` generates fatal failures. With expect you can see all the conditions that fail and assert you can only see the first. Normally assert is used to mark a condition that would invalidate follow on conditions if failed

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
```diff
@@ -0,0 +1,607 @@
...
+    {
+        _running = false;
+        if (_exchangeThread.joinable()) {
+            _exchangeThread.join();
+        }
+    }
+
+    void PriorityPlugin::OnStateChange(IvpPluginState state)
+    {
+        OnStateChange(state);
```
paulbourelly999: I think  you mean to call the parent function not make  a recursive call here. I think these needs to be 
```
PluginClientClockAware::OnStateChange(state)
```

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
```diff
@@ -0,0 +1,607 @@
...
+                _running = true;
+                _exchangeThread = std::thread(&PriorityPlugin::ServiceExchangeLoop, this);
+                PLOG(logINFO) << "Started PRS-CO service exchange loop thread.";
+            }
+		}
+    }
+
+    void PriorityPlugin::OnConfigChanged(const char *key, const char *value)
+    {
+        OnConfigChanged(key, value);
```
paulbourelly999: Same comment as above
```
PluginClientClockAware::OnConfigChanged(key,value);
```

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
```diff
@@ -0,0 +1,607 @@
...
+                try {
+                    info.snmpClient = std::make_shared<snmp_client>(
+                        info.ip, info.port, _snmpCommunity,
+                        "", "", "", "", "", "",
+                        SNMP_VERSION_1, SNMP_DEFAULT_TIMEOUT);
+                } catch (const snmp_client_exception &e) {
+                    PLOG(logERROR) << "Failed to create SNMP client for IntersectionID="
+                                   << info.intersectionID << " (" << info.ip << ":" << info.port
+                                   << "): " << e.what();
+                    continue;
```
paulbourelly999, line 147: Do we want to continue here if we fail to create a SNMP client. At minimum lets make a UI notification that something failed. Either a plugin status to indicate this connection failed or a TmxEventLog or both

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
```diff
@@ -0,0 +1,607 @@
...
+            auto timeOfRequest = static_cast<uint32_t>(nowEpoch);
+
+            if (auto existing = _requestorStates.find(vehicleKey);
+                existing != _requestorStates.end() && existing->second.sequenceNumber == newSeq) {
+                PLOG(logDEBUG1) << "SRM sequence number unchanged (" << static_cast<int>(newSeq)
+                                << ") for this vehicle, discarding.";
+                return;
+            }
+
+            SweepStaleTrackedRequests(std::chrono::steady_clock::now());
```
paulbourelly999: Use plugin clock

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
```diff
@@ -0,0 +1,607 @@
...
+            PLOG(logERROR) << "Invalid TSC_Controllers entry: " << e.what();
+        }
+
+        PLOG(logINFO) << "PriorityPlugin configured: " << _controllers.size() << " controller(s)"
+                                    << " Role=" << _pluginRole
+                                    << " PollInterval=" << _pollIntervalMs << "ms"
+                                    << " TimeToLive=" << _timeToLiveSec << "s";
+    }
+
+    void PriorityPlugin::HandleSRM(SrmMessage &msg, tmx::routeable_message &routeableMsg)
```
paulbourelly999, line 168: If we can move more of the logic in this function out of our plugin into a separate class function or free function we can unit test this. The only thing we should not be able to move out is any plugin functionality like status or message broadcast

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
```diff
@@ -0,0 +1,607 @@
...
+                                (classType == other.vehicleClassType && classLevel < other.vehicleClassLevel);
+            if (isActive && isLowerClass) {
+                other.statusInPRS = RequestStatus::activeOverride;
+                PLOG(logINFO) << "New request overrode active entry in slot: " << (&other - &table[0])
+                              << " with lower priority class. Marking overridden entry as activeOverride.";
+            }
+        }
+    }
+
+    void PriorityPlugin::ProcessPrgPackage(const SignalRequestPackage &pkg,
```
paulbourelly999: If we can move more of this functionality out of this plugin function, into a separate class function or free function we can cover more of it with unit tests. The only thing we should not be able to move out are plugin functionality like setting plugin status or broadcasting a message

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.hpp
```diff
@@ -0,0 +1,250 @@
...
+            void HandleSRM(tmx::messages::SrmMessage &msg, tmx::routeable_message &routeableMsg);
+
+        private:
+            // PRG: request tracker 
+            // Tracks requests sent to the PRS so we can track new requests, updates, and when to send clear after cancel
+            enum class PrgRequestState : uint8_t {
+                sent,
+                canceled
+            };
+
```
paulbourelly999: NIT: Missing Comment

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.hpp
```diff
@@ -0,0 +1,250 @@
...
+             * @brief Handles incoming Signal Request Messages (SRMs).
+             * @param msg the SRM received.
+             * @param routeableMsg the original routeable message that was received.
+             */
+            void HandleSRM(tmx::messages::SrmMessage &msg, tmx::routeable_message &routeableMsg);
+
+        private:
+            // PRG: request tracker 
+            // Tracks requests sent to the PRS so we can track new requests, updates, and when to send clear after cancel
+            enum class PrgRequestState : uint8_t {
```
paulbourelly999: NIT: We are defining these as private classes inside our plugin C++ file. This means no other classes should have access to these structs. This forces us  to have a lot of the logic handling these structs in our plugin class which cannot be covered by unit tests. I suggest making each of these their on hpp and if necessary cpp files. This makes it easier at a glance to see what data types exists. It also makes it easier to just include the data types you need. Just a suggestion

src/v2i-hub/PriorityPlugin/src/PriorityPluginMessaging.cpp
paulbourelly999: This seems to still be missing a push

src/v2i-hub/PriorityPlugin/src/PriorityPluginMessaging.cpp
```diff
@@ -0,0 +1,539 @@
...
+            ASN_STRUCT_FREE_CONTENTS_ONLY(asn_DEF_SignalStatusMessage, ssmPtr.get());
+
+            encodedSSM.set_flags(IvpMsgFlags_RouteDSRC);
+            encodedSSM.addDsrcMetadata(0xE0000015);
```
paulbourelly999: NIT: Should we be using the constants you update for this

src/v2i-hub/PriorityPlugin/src/PriorityPluginMessaging.cpp
paulbourelly999, line 1: The functions defined in this CPP are declared in the PriorityPlugin HPP file and are functions of the priority plugin. Separating this into its own uniquely named CPP that does not correspond with any class or free function just makes this harder to read since it is not intuitive to determine where the functions in the priority plugin are define or inversely to which class these functions belong. It also provides not benefit in terms of unit testing since this is all still not unit testable. My suggestion for logic like this is to handle the logic in a free function or separate class function that can be unit tested and just handle the plugin functionality in the plugin which cannot be tested. Something like this 

```CPP
class PriorityPlugin :: public tmx::utils::PluginClientClockAware {
   public:
      void HandleSRM(tmx::messages::SrmMessage &msg, tmx::routeable_message &routeableMsg) {
          // do core logic in separate free function or class function
         PLOG(logINFO) << "Received Signal Request Message (SRM)";
         tmx::messages::SrmEncodedMessage srmEnc(routeableMsg);
          auto srmDecoded = srmEnc.decode_j2735_message();
          auto srm = srmDecoded.get_j2735_data();
          if (srm) {
              if (!srm->requests || srm->requests->list.count <= 0) {
                PLOG(logWARNING) << "SRM contains no signal request packages, skipping.";
                _skippedMessages++;
                SetStatus(_keySkippedMessages, _skippedMessages);
            }
          } else {
          try {
               tmx::messages::SsmEncodeMessage ssm = PriorityPlugin::CoreLogic::ProcessSRM( ...)
               BroadcastMessage(ssm);
          } catch ( const UnchangedSequenceNumberException &e) {
              PLOG(logERROR) << " Unchanged sequence number exception :  " << e.what();
              SetStatus(_keySkippedMessages, _skippedMessages);
            }
          }
         }

          
      };

}
```

The key things I am suggesting here are, using exceptions or return types in methods to break out of regular function control instead of empty return statements. Moving core logic like processing requests and generating SRMs into separate function that is unit testable. Trying to limit logic in Plugin to Broadcast and Status changes

src/v2i-hub/PriorityPlugin/src/PriorityPluginMessaging.cpp
jwillmartin: Thanks, just pushed

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
jwillmartin: Updated

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
jwillmartin: Good catch, updated

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
jwillmartin: Updated

src/tmx/TmxApi/tmx/TmxApiMessages.h
```diff
@@ -238,7 +238,7 @@ static CONSTEXPR const char *STATUS_STOPPED_STRING = "Stopped / Disconnected";
...
+	mapData_PSID = 0xE0000017,
```
dan-du-car: Can we add a note why these new PSID is chosen and provide any references if applicable.

src/tmx/TmxUtils/src/tsc/NTCIP_1211_MIB.h
```diff
@@ -0,0 +1,25 @@
...
+    static const std::string NTCIP1211_PRIORITY_UPDATE_ABSOLUTE_OID = "1.3.6.1.4.1.1206.4.2.11.2.9.0";
+
+    // prgPriorityCancel (1211 v02A-SE03f PRS-MIB1 5.1.2.5)
+    static const std::string NTCIP1211_PRIORITY_CANCEL_OID = "1.3.6.1.4.1.1206.4.2.11.2.5.0";
+
+    // prgPriorityClear (1211 v02A-SE03f PRS-MIB1 5.1.2.6)
+    static const std::string NTCIP1211_PRIORITY_CLEAR_OID = "1.3.6.1.4.1.1206.4.2.11.2.6.0";
+
+    // prsServiceRequest (1211 v0224j CO-MIB1 5.2.2.1)
+    static const std::string NTCIP1211_PRS_SERVICE_REQUEST_OID = "1.3.6.1.4.1.1206.4.2.11.4.1.0";
```
dan-du-car: Nit: We could remove the repeat of NTCIP1211 as the namespace has already stated ntcip1211

src/v2i-hub/PriorityPlugin/scripts/mock-srm.py
```diff
@@ -0,0 +1,600 @@
...
+    "priorityCancellation"
+]
+
+
+def get_moy() -> int:
+    """Minute-of-year for the current UTC time."""
+    now = datetime.datetime.now(tz=datetime.timezone.utc)
+    start_of_year = datetime.datetime(now.year, 1, 1, tzinfo=datetime.timezone.utc)
+    delta = now - start_of_year
+    return delta.days * 1440 + now.hour * 60 + now.minute
```
dan-du-car: Nit: using meaningful variables for 1440 and 60 and 1000

src/v2i-hub/PriorityPlugin/scripts/mock-srm.py
```diff
@@ -0,0 +1,600 @@
...
+
+
+def get_d_second() -> int:
+    """Millisecond within the current UTC minute."""
```
dan-du-car: The function seems to get the milliseconds from current seconds and microseconds rather than minutes stated in the comment.

src/v2i-hub/PriorityPlugin/scripts/mock-srm.py
```diff
@@ -0,0 +1,600 @@
...
+
+        self.intersection_id = QSpinBox()
+        self.intersection_id.setRange(0, 65535)
+        self.intersection_id.setValue(defaults.get("intersection_id", 9709))
+
+        self.inbound_lane = QSpinBox()
+        self.inbound_lane.setRange(0, 255)
+        self.inbound_lane.setValue(defaults.get("inbound_lane", 1))
+
+        self.distance = QDoubleSpinBox()
```
dan-du-car, line 245: Nit: Could we add title with short description to explain the terms and unit of measure used: distance_m: distance to intersection in unit of meter?

src/v2i-hub/PriorityPlugin/scripts/mock-srm.py
```diff
@@ -0,0 +1,600 @@
...
+            "entity_id": self.entity_id.text(),
+            "role": self.role.currentText(),
+            "request_type": self.request_type.currentText(),
+        }
+
+
+class MainWindow(QMainWindow):
+    def __init__(self):
+        super().__init__()
+        self.setWindowTitle("Mock SRM Sender")
```
dan-du-car: Nit: Spell out Signal Request Message (SRM)

src/v2i-hub/PriorityPlugin/scripts/requirements.txt
```diff
@@ -0,0 +1 @@
+PyQt6
```
dan-du-car: add a specific version

src/v2i-hub/PriorityPlugin/src/PriorityConfiguration.cpp
```diff
@@ -0,0 +1,106 @@
...
+            return configs;
+        }
+
+        boost::property_tree::ptree pt;
+        std::istringstream iss(json);
+        boost::property_tree::read_json(iss, pt);
+
+        for (const auto &[_, node] : pt) {
+            ControllerConfig cfg;
+            cfg.intersectionID = node.get<long>(IntersectionIDKey);
```
dan-du-car, line 46: Nit: Do we want to add exception logic when the keys do not exist in the json?

src/v2i-hub/PriorityPlugin/src/PriorityConfiguration.cpp
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        boost::property_tree::ptree pt;
+        std::istringstream iss(json);
+        boost::property_tree::read_json(iss, pt);
+
+        for (const auto &[_, node] : pt) {
+            auto intID       = node.get<long>(IntersectionIDKey);
+            auto lane        = node.get<long>(LaneKey);
+            auto strategyVal = node.get<int>(StrategyKey);
```
dan-du-car, line 68: The same comment as the above

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
```diff
@@ -0,0 +1,481 @@
...
+            long role = 0;
+            if (srm->requestor.type != nullptr) {
+                role = srm->requestor.type->role;
+            }
+            auto [classType, classLevel] = MapVehicleClass(role);
+
+            uint8_t newSeq = srm->sequenceNumber ? static_cast<uint8_t>(*srm->sequenceNumber) : 0;
+
+            // PRS mode: handle SRM per NTCIP 1211 4.2.3.1 (how PRS receives a priority request from a PRG).
+            if (_pluginRole == "PRS") {
```
dan-du-car: What if the _pluginRole is not PRS nor PRG. Did I miss a validation logic for the role?

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.hpp
```diff
@@ -0,0 +1,189 @@
...
+
+            // Map of request ID to tracked PRG request for active requests
+            std::unordered_map<std::string, PrgTrackedRequest> _prgTrackedRequests;
+
+            // PRS mode: last seen SRM sequence number per vehicle, used to drop
+            // duplicates delivered by multiple intersections near the vehicle.
+            std::unordered_map<std::string, uint8_t> _prsLastSeqByVehicle;
+
+            // For PRS mode: NTCIP 1211 priority request processor.
+            PriorityRequestProcessor _processor;
```
dan-du-car, line 150: Is this _processor initialized before used in the .cpp file?

src/tmx/TmxApi/test/J2735MessageTest.cpp
jwillmartin: Updated to use the macro

src/tmx/TmxApi/test/J2735MessageTest.cpp
jwillmartin: Good call, replaced all

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
jwillmartin: Updated to add error to UI. It should continue because there could be other connected controllers.

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
jwillmartin: Updated to move majority of logic to PriorityPluginWorker and added tests. Moved more logic to SsmBuilder and PriorityRequestBuilder and added unit tests.

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
jwillmartin: Moved majority out. Let me know what you think

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.hpp
jwillmartin: Moved to PriorityTypes.hpp, added comment.

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.hpp
jwillmartin: Good call, moved to PriorityTypes.hpp. Other classes also declared in other appropriate header files

src/v2i-hub/PriorityPlugin/src/PriorityPluginMessaging.cpp
jwillmartin: Good catch, updated

src/v2i-hub/PriorityPlugin/src/PriorityPluginMessaging.cpp
jwillmartin: Moved most logic to new files. Now in PriorityRequestBuilder/Processor and PrsSerivceExchange.

src/tmx/TmxApi/tmx/TmxApiMessages.h
jwillmartin: Added note

src/tmx/TmxUtils/src/tsc/NTCIP_1211_MIB.h
jwillmartin: Good idea, removed

src/v2i-hub/PriorityPlugin/scripts/requirements.txt
jwillmartin: Added version

src/v2i-hub/PriorityPlugin/scripts/mock-srm.py
jwillmartin: Added consts

src/v2i-hub/PriorityPlugin/scripts/mock-srm.py
jwillmartin: Had to build it using the current microsecond in the second and current second in the minute. But updated to use more precise absolute nanoseconds since epoch.

src/v2i-hub/PriorityPlugin/scripts/mock-srm.py
jwillmartin: Updated

src/v2i-hub/PriorityPlugin/scripts/mock-srm.py
jwillmartin: The description is set later when we add rows to the layout. The unit is also added for each of these as a suffix

src/v2i-hub/PriorityPlugin/src/PriorityConfiguration.cpp
jwillmartin: Updated to be similar to the configuration exceptions in the RSUHealthMonitorPlugin

src/v2i-hub/PriorityPlugin/src/PriorityConfiguration.cpp
jwillmartin: Same, updated. Also updated tests.

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
jwillmartin: Updated to handle this during configuration and added tests

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.hpp
jwillmartin: I believe it's initialized when the plugin is constructed, so it's available before it's used

src/tmx/TmxApi/test/J2735MessageTest.cpp
paulbourelly999: Am I missing something here. It still looks like a regular free call to me.

src/v2i-hub/PriorityPlugin/scripts/install.sh
paulbourelly999, line 1: Can we add a comment here or name this more descriptively that it is installing requirements for mock-srm.py. I know we have the README but people going through our repository looking for scripts may accidently believe this is a script to install the priority plugin.

src/v2i-hub/PriorityPlugin/scripts/README.md
```diff
@@ -0,0 +1,19 @@
+# Mock-SRM
+
+The [mock-srm.py](mock-srm.py) script allows for testing multiple SRMs with the Priority Plugin.
```
paulbourelly999, line 3: NIT: If we can describe in more detail what this script is actually mocking that would be amazing. Is it hosting a SNMP server? What parts of the NTCIP 1211 functionality is it providing.

src/v2i-hub/PriorityPlugin/src/PrsServiceExchange.hpp
```diff
@@ -0,0 +1,82 @@
...
+    };
+
+    /**
+     * @brief Sends the encoded OCTET STRING to a TSC via SNMP SET.
+     * @param client the SNMP client for the target controller
+     * @param oidStr the NTCIP 1211 OID to SET
+     * @param data the OER-encoded payload
+     * @return true on success, false on failure or if the client is null.
+     */
+    bool SnmpSet(const std::shared_ptr<tmx::utils::snmp_client> &client, const std::string &oidStr, const std::vector<uint8_t> &data);
```
paulbourelly999, line 60: NIT: In the future we may want to move these methods into our existing SNMP Client class. Having some SNMPSet/SNMPGet functions hidden in a different plugin may cause others how need similar functionality to recreate it instead of reusing it

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
```diff
@@ -0,0 +1,494 @@
...
+        PLOG(logINFO) << "PRS service exchange loop stopped.";
+    }
+
+    void PriorityPlugin::EncodeAndBroadcastSSM(const std::shared_ptr<SignalStatusMessage_t> &ssmPtr)
+    {
+        try {
+            SsmEncodedMessage encodedSSM;
+            MessageFrameMessage frame(ssmPtr);
+            encodedSSM.set_data(TmxJ2735EncodedMessage<SignalStatusMessage>::encode_j2735_message<codec::uper<MessageFrameMessage>>(frame));
+            free(frame.get_j2735_data().get()); // NOSONAR: ASN.1 C struct allocated via calloc
```
paulbourelly999: We should remove any calls to free the C Structs inside of the TmxJ2735EncodedMessages as well as the TmxJ2735Messages because we are working to make those classes manage the raw pointers themselves. See this line, which I temporarily had to comment out due to a heap-use-after-free segmentation fault (https://github.com/usdot-fhwa-OPS/V2X-Hub/blob/7554c7703fb22ebba6ecd8d0874277f2adca26fe/src/tmx/TmxApi/tmx/messages/TmxJ2735.hpp#L98).

src/tmx/TmxApi/test/J2735MessageTest.cpp
paulbourelly999: Also see this comment We should remove any calls to free the C Structs inside of the TmxJ2735EncodedMessages as well as the TmxJ2735Messages because we are working to make those classes manage the raw pointers themselves. See this line, which I temporarily had to comment out due to a heap-use-after-free segmentation fault (https://github.com/usdot-fhwa-OPS/V2X-Hub/blob/7554c7703fb22ebba6ecd8d0874277f2adca26fe/src/tmx/TmxApi/tmx/messages/TmxJ2735.hpp#L98).

src/v2i-hub/PriorityPlugin/src/SsmBuilder.cpp
```diff
@@ -0,0 +1,335 @@
...
+
+        // Allocates a zeroed byte buffer for an ASN.1 OCTET STRING.
+        uint8_t *AllocAsnBuffer(size_t size) {
+            return static_cast<uint8_t *>(calloc(size, 1)); // NOSONAR - freed via free() by ASN_STRUCT_FREE
+        }
+
+        // Allocates a SignalStatusMessage_t, owned by a shared_ptr.
+        std::shared_ptr<SignalStatusMessage_t> MakeOwnedSsm() {
+            return {AllocAsn<SignalStatusMessage_t>(),
+                    [](SignalStatusMessage_t *p) { j2735::j2735_destroy<SsmTraits>(p); }};
```
paulbourelly999, line 60: I think SsmTraits is the wrong type descriptor here

src/v2i-hub/PriorityPlugin/src/SsmBuilder.cpp
```diff
@@ -0,0 +1,335 @@
...
+        template <typename T>
+        T *AllocAsn(T value) {
+            auto *p = AllocAsn<T>();
+            *p = value;
+            return p;
+        }
+
+        // Allocates a zeroed byte buffer for an ASN.1 OCTET STRING.
+        uint8_t *AllocAsnBuffer(size_t size) {
+            return static_cast<uint8_t *>(calloc(size, 1)); // NOSONAR - freed via free() by ASN_STRUCT_FREE
```
paulbourelly999: Lets please remove any NOSONAR comments. If we can not remove calloc statements in our current implementation we still want sonar to report and keep track of them

src/v2i-hub/PriorityPlugin/src/SsmBuilder.cpp
```diff
@@ -0,0 +1,335 @@
...
+    namespace {
+        // Note for allocation helpers below: 
+        // ASN.1 allocations must come from the malloc family.
+        // j2735_destroy releases the SSM tree with ASN_STRUCT_FREE, which calls free() on every node.
+        // calloc's zeroing also marks all optional fields absent.
+
+        // Allocates a zeroed ASN.1 struct.
+        template <typename T>
+        T *AllocAsn() {
+            return static_cast<T *>(calloc(1, sizeof(T))); // NOSONAR - freed via free() by ASN_STRUCT_FREE
```
paulbourelly999: Lets please remove any NOSONAR comments. If we can not remove calloc statements in our current implementation we still want sonar to report and keep track of them

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
paulbourelly999: Also Lets please remove any NOSONAR comments. If we can not remove calloc statements in our current implementation we still want sonar to report and keep track of them

src/v2i-hub/PriorityPlugin/src/PriorityPluginWorker.cpp
```diff
@@ -0,0 +1,115 @@
...
+            case 11: return {1, 7}; // roadSideSource
+            case 1:  return {3, 1}; // publicTransport
+            case 16: return {3, 2}; // transit
+            case 15: return {5, 1}; // dot
+            case 4:  return {5, 2}; // roadWork
+            case 3:  return {7, 1}; // dangerousGoods
+            case 9:  return {7, 2}; // truck
+            case 17: return {7, 3}; // slowMoving
+            case 18: return {7, 4}; // stopNgo
+            default: return {10, 1};
```
paulbourelly999: What is the default case?

src/v2i-hub/PriorityPlugin/src/PrsServiceExchange.hpp
jwillmartin: I was thinking of doing this, but noticed other plugins also built their own set/get functions, so I kept it here. We can create a story to move this to the snmp client.

src/v2i-hub/PriorityPlugin/src/PriorityPlugin.cpp
jwillmartin: Good catch, removed the free.

src/v2i-hub/PriorityPlugin/scripts/README.md
jwillmartin: updated readme

src/v2i-hub/PriorityPlugin/scripts/install.sh
jwillmartin: Updated with comment

src/v2i-hub/PriorityPlugin/src/SsmBuilder.cpp
jwillmartin: Verified it's correct

src/v2i-hub/PriorityPlugin/src/SsmBuilder.cpp
jwillmartin: Removed

src/v2i-hub/PriorityPlugin/src/SsmBuilder.cpp
jwillmartin: Removed

src/v2i-hub/PriorityPlugin/src/PriorityPluginWorker.cpp
jwillmartin: Anything that typically wouldn't request priority. Added comment to clarify
Set them to the lowest class type and level.

src/tmx/TmxApi/test/J2735MessageTest.cpp
jwillmartin: Good catch, made changes and tested. Had to temporarily comment out other tests and roadsafetymessage.cpp to test for leaks from the two new tests introduced here. Updates didn't result in any ASan errors.

paulbourelly999 (changes requested):
Final comment, one of our integration test cases in CI is failing, I think it maybe related to one of your changes. Lets try to fix this tomorrow before we merge

paulbourelly999 (approved):

PR #885, opened 2026-05-07
Updated CARMACloudPlugin to use HTTPS

kelleydi:
# PR Details
## Description

V2X-Hub and CARMA Cloud currently communicate over HTTP through SSH tunnels. This change updates CARMACloudPlugin to use HTTPS over the existing SSH tunnel infrastructure and addresses SonarQube findings related to deprecated SSL/TLS configurations.

### Specific Changes
* Updated `CloudSend` to use HTTPS communication
* Updated `CloudSendAsync` to capture variables by value
  * **Reason:** The detached thread previously captured `local_msg`, `local_url`, `local_base`, and `local_method` by reference. These references could become invalid after the function returned, potentially causing undefined behavior or stability issues.
* Simplified manifest configurations to use `CARMACloudBaseUrl` and set default to `https://host.docker.internal:33333` 
* Added `enforceTLSVerification`  and  `carma_cloud_ca_cert_path` configurations support to `CARMACloudPlugin` and exposed it through the manifest
* Updated SSH tunnel configuration to enable HTTPS traffic from V2X-Hub to CARMA-Cloud:
  * used the `-g` flag which permits non-local connections to the forwarded port, enabling Docker containers to access the tunnel endpoint established on host.
  * forwarded HTTPS traffic over port `8443` 
* Updated documentation to clarify plugin configuration options and explain when `host.docker.internal` should be used

## Related Jira Key
[ OMDO-158](https://usdot-carma.atlassian.net/browse/OMDO-158)

## Motivation and Context
 It addressed SonarQube finding to use stronger SSL and TLS versions.

## How Has This Been Tested?
This was tested manually.

### Test Steps
- Build and deploy V2X Hub
- Enable Message Receiver and CARMA Cloud Plugins
- Configure CARMA Cloud Plugin for CARMA Cloud connection
- Establish SSH tunneling from host
- Use [tcr_script.py](https://github.com/usdot-fhwa-OPS/V2X-Hub/blob/develop/src/v2i-hub/CARMACloudPlugin/scripts/tcr_script.py) to send a Traffic Control Request (TCR) to the Message Receiver which will forward to CARMA Cloud Plugin which sends it to CARMA Cloud.
- Verify V2X-Hub logs (with DEBUG enabled) show TCR message going from Message Receiver → CARMA Cloud Plugin → CARMA Cloud
- Verify CARMA Cloud logs received request.
[...]

PR #891, opened 2026-05-13
Add SSL for ERV Message Forwarding CARMA Cloud connection

paulbourelly999:
# PR Details
## Description

Follow on for #885 to update ERV Message forwarding with TLS for V2X Hub -> CARMA Cloud connection.
**Note**: Several improvements should be made in the future to this implementation
1) Since this plugin connects to CARMA Cloud, the functionality for ERV should be moved to the CARMA Cloud Plugin to reduce duplication
2) The duplicate SNMP Client in this plugin is removed by https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/939 due to memory leaks in this implementation and to remove duplication
3) This plugin should use RSUStatusMessage to provide updates to CARMA Cloud on RSU location instead of implementing duplicate RSU SNMP OID polling

Since we do not have enough scope in the current release to make these changes and test them, this will be left for future changes.
## Related Jira Key
[OMDO-158
](https://usdot-carma.atlassian.net/browse/OMDO-158)

## Motivation and Context
Improve security for V2X Hub

## How Has This Been Tested?
Tested by 
1) commenting out registerRsuLocation calls to prevent SNMP failures in local testing 
2) Sending the following BSM data to the message receiver `00143d604043030280ffdbfba868b3584ec40824646400320032000c888fc834e37fff0aaa960fa0040d082408801148d693a431ad275c7c6b49d9e8d693b60e`
3) Confirming HTTPS REST calls to post this data could be made to carma cloud development server
image

## Types of changes

## Checklist:

### DIFF
--- a/src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.cpp
+++ b/src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.cpp
@@ -69,7 +69,7 @@ namespace ERVCloudForwardingPlugin
                 string rsu_identifier = _rsuName + "_" + boost::lexical_cast<std::string>(uuid);
                 auto xml_str = ERVCloudForwardingWorker::constructRSULocationRequest(rsu_identifier, _webPort, latitude, longitude);
                 PLOG(logINFO) << "Sending registering RSU location reqest to cloud: " << xml_str << endl;
-                auto status = CloudSend(xml_str, _CLOUDURL, _CLOUDRSUREQ, _POSTMETHOD);
+                auto status = CloudSend(xml_str, carma_cloud_url, _CLOUDRSUREQ, _POSTMETHOD);
                 if (status == 1)
                 {
                     PLOG(logERROR) << "Cannot register RSU location. Reason: Failed to send RSU location to cloud." << endl;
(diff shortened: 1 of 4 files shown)


src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.cpp
```diff
@@ -182,30 +191,66 @@ namespace ERVCloudForwardingPlugin
...
+        if (!enforceTLSVerification) {
+            curl_easy_setopt(req, CURLOPT_SSL_VERIFYPEER, 0L);
+            curl_easy_setopt(req, CURLOPT_SSL_VERIFYHOST, 0L);
+            PLOG(logWARNING) << "TLS verification disabled by setting configuration 'enforceTLSVerification' to false. "
+                << "CARMA-Cloud certificate and hostname validation are NOT being enforced. "
+                << "This option should only be disabled in non-production / development environments for temporary troubleshooting.";
+        }
+        // END-NOSCAN
+        #else
+        if (!enforceTLSVerification) {
```
dan-du-car: Can these two same !enforceTLSVerification if statement checks be combined?

src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.cpp
```diff
@@ -182,30 +191,66 @@ namespace ERVCloudForwardingPlugin
...
-                    fprintf(stderr, "curl send failed: %s\n", curl_easy_strerror(res));
-                    return EXIT_FAILURE;
-                }else{
-                    PLOG(logDEBUG) << "Successfully forwarded message to cloud via curl: " << local_msg << endl;
-                }
-            }
+        CURL* req = curl_easy_init();
+        if (!req) return 1;
+
+        const std::string urlfull = local_url + local_base;
```
dan-du-car, line 197: Should we consider a separate function to create the urlful string with validation of the local_url and local_base to ensure the output of this urlful is following correct URL format?

dan-du-car:
Does bullet point 3 in the PR description still apply?

src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.cpp
paulbourelly999: I have updated to include both inside the same if statement but look carefully. Because of the macros, only one of these ever actually gets compiled. That is the point. To allow during debug builds where ALLOW_INSECURE_TLS is set to true, developers to test the communication without HTTPS but to remove it from all released images

src/v2i-hub/ERVCloudForwardingPlugin/src/ERVCloudForwardingPlugin.cpp
paulbourelly999: I think this is a one line operation and is probably fine without a separate function. If in the future we decide to do more verification on the URL creation we could put that in a separate method but that is not in the scope for this story

dan-du-car (approved):

PR #907, opened 2026-06-22
dt4it-105: Added custom Message and Unit tests

mpalash95:
# PR Details
## Description

1. Added new  TMX message SecuredV2XMessage  to support signed V2X messages
2. Added Unit tests for newly added message

## Related Jira Key
[DT4IT-105](https://usdot-carma.atlassian.net/jira/software/c/projects/DT4IT/boards/591/backlog?selectedIssue=DT4IT-105)

## How Has This Been Tested?
Added unit tests and tested by running test task in dev container 

## Types of changes

## Checklist:

[DT4IT-105]: https://usdot-carma.atlassian.net/browse/DT4IT-105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/tmx/Messages/include/SecuredV2XMessage.h
```diff
@@ -0,0 +1,61 @@
...
+#include <vector>
+#include <cstdint>
+#include "MessageTypes.h"
+namespace tmx {
+namespace messages {
+
+class DsrcMetadata : public tmx::message
+{
+public:
+    DsrcMetadata() {}
```
paulbourelly999: Why do we need this new class. DSRC metadata can already be passed with TmxMessages on broadcast. Also this information should not be relevant to the HWIL setup since it should be dropped and replaced outside of the proxy radios

src/tmx/Messages/include/SecuredV2XMessage.h
```diff
@@ -0,0 +1,61 @@
...
+    DsrcMetadata(const tmx::message_container_type &contents): tmx::message(contents) {}
+
+    /// DSRC PSID.
+    std_attribute(this->msg, int, psid, 0, )
+    /// DSRC channel.
+    std_attribute(this->msg, int, channel, 0, )
+};
+
+class SecuredV2XMessage : public tmx::message
```
paulbourelly999: Lets rename this message. Technically it does not need to be a secure message. Really it is just the raw full spdu bytes. Something like RawSpdu or Spdu or whatever the acronym stands for or Raw1602dot2Message all seem like better alternatives to me

paulbourelly999: Please document this class. What is this message for. What fields need to be populated and how. UUID, timestamp etc

src/tmx/Messages/include/SecuredV2XMessage.h
```diff
@@ -0,0 +1,61 @@
...
+std_attribute(this->msg, std::string, uuid, "", )
+    /// J2735 payload type, e.g. "BSM", "PSM", "MAP".
+std_attribute(this->msg, std::string, MessageType, "", )
+    /// Packet reception time, epoch milliseconds.
+std_attribute(this->msg, int64_t, timestamp_ms, 0, )
+    /// DSRC metadata: PSID and channel.
+std_attribute(this->msg, DsrcMetadata, dsrc_metadata, DsrcMetadata(), )
+
+    // Smart-pointer accessors for spdu_data
+void set_spdu_data_bytes(std::shared_ptr<std::vector<uint8_t>> bytes) {
```
paulbourelly999: Why does this need to be a pointer at all?

paulbourelly999: Also please see the https://github.com/usdot-fhwa-OPS/V2X-Hub/blob/c294ab147fc4f19379e4326c592539d14e81b544/src/tmx/TmxApi/tmx/messages/byte_stream.hpp#L23 `byte_stream` typedef and to/from string conversion methods. Please use this instead

src/tmx/Messages/include/SecuredV2XMessage.h
```diff
@@ -0,0 +1,61 @@
...
+std::shared_ptr<std::vector<uint8_t>> get_spdu_data_bytes() {
+std::string s = get_spdu_data();
+return std::make_shared<std::vector<uint8_t>>(s.begin(), s.end());
+    }
+    // Smart-pointer accessors for uuid
+void set_uuid_bytes(std::shared_ptr<std::vector<uint8_t>> bytes) {
+set_uuid(std::string(bytes->begin(), bytes->end()));
+    }
+std::shared_ptr<std::vector<uint8_t>> get_uuid_bytes() {
+std::string s = get_uuid();
```
paulbourelly999: Please fix the formatting in this file. It is very hard to read without proper indentation

src/tmx/Messages/test/SignedMessageTest.cpp
```diff
@@ -0,0 +1,110 @@
...
+#include "SecuredV2XMessage.h"
+#include <vector>
+#include <iostream>
+
+using namespace std;
+using namespace tmx;
+using namespace tmx::messages;
+
+// Function to convert hex string to byte array
+std::vector<uint8_t> hexStringToByteArray(const std::string& hex) {
```
paulbourelly999: See comment about byte_stream

paulbourelly999 (changes requested):

src/tmx/Messages/include/SecuredV2XMessage.h
mpalash95: Removed DSRCMetaData Class

src/tmx/Messages/include/SecuredV2XMessage.h
mpalash95: Renamed to RawSpdu  and added comment about the class

src/tmx/Messages/include/SecuredV2XMessage.h
mpalash95: Added byte_stream fields to RawSPDU

src/tmx/Messages/include/SecuredV2XMessage.h
mpalash95: Fixed Formatting

src/tmx/Messages/test/SignedMessageTest.cpp
mpalash95: Addressed

src/tmx/Messages/include/SecuredV2XMessage.h
mpalash95: > Why does this need to be a pointer at all?

design_doc

Created shared Pointer for SPDU as per the design document

src/tmx/Messages/include/RawSpdu.h
```diff
@@ -0,0 +1,59 @@
...
+    namespace messages
+    {
+        /**
+         * RawSpdu - TMX message wrapper carrying a raw, encoded SPDU.
+         *
+         * Holds an already-encoded 1609.2-secured J2735 message, such as BSM, PSM,
+         * MAP, SPaT, etc., to move raw V2X payloads across the TMX bus.
+         *
+         * Fields (none are set automatically; the caller must populate them before publishing):
+         *   - spdu_data:    Raw SPDU bytes (std::string is used as a byte buffer).
```
kjrush: Is this storage as `std::string` being handled at the TMX message layer, as in the implementation of `tmx::byte_stream` is doing this? I think my overall preference would be to treat this as a vector of uint8s as much as possible, but if the underlying message framework prefers to treat things as strings I think that's acceptable.

src/tmx/Messages/include/RawSpdu.h
```diff
@@ -0,0 +1,59 @@
...
+         *                   via set_spdu_data_bytes()/get_spdu_data_bytes().
+         *   - uuid:         Unique ID for this packet, used for correlation/tracing
+         *                   (e.g., a UUID v4 string). Use set_uuid()/get_uuid().
+         *   - MessageType:  J2735 type label, e.g., "BSM", "PSM", or "MAP".
+         *   - timestamp_ms: Epoch timestamp in milliseconds, received or generated time
+         *                   depending on the producer. Defaults to 0 if unset.
+         */
+        class RawSpdu : public tmx::message
+        {
+        public:
```
kjrush, line 36: I think this message type also needs the DSRC metadata fields, unless we're pulling that in via the inheritance from `tmx::message`. Can you confirm or add these fields as needed? We'll need these on the outgoing side I believe, to know which PSID at least to request the ProxyRadio to transmit with.

 Image

kjrush: After discussion w/ Paul, the use of this specific dsrc_metadata struct might not be appropriate, since that already has some kind of meaning in the system and could cause confusion in the future. But adding fields here or defining a new struct to capture these same PSID and channel parameters would be good still, maybe a new 1609Dot2Metadata struct or something.

src/tmx/Messages/include/RawSpdu.h
kjrush: It sounds like this string comment is from an earlier version, after talking with Paul. Lets update this to reflect current implementation.

src/tmx/Messages/include/RawSpdu.h
mpalash95: added as  fileds in the message

src/tmx/Messages/include/RawSpdu.h
mpalash95: Updated the comment

kjrush (dismissed):

src/tmx/Messages/test/RawSpduTest.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include "RawSpdu.h"
+#include <vector>
+#include <iostream>
+
+using namespace std;
+using namespace tmx;
+using namespace tmx::messages;
+
+// Function to convert hex string to byte array
+tmx::byte_stream hexStringToByteArray(const std::string& hex) {
```
paulbourelly999: This function should no longer be necessary if using the byte_stream type

src/tmx/Messages/test/RawSpduTest.cpp
paulbourelly999, line 1: Can we add a test case to see how the message is serialized to string. Simply create a message and call to_string() on it and see if the JSON results correct converts byte_stream to string

paulbourelly999 (changes requested):
Small changes to unit test

src/tmx/Messages/test/RawSpduTest.cpp
mpalash95: Removed the function

src/tmx/Messages/test/RawSpduTest.cpp
mpalash95: Added a unit test to test JSON result

src/tmx/Messages/include/RawSpdu.h
```diff
@@ -0,0 +1,66 @@
...
+         *   - psid:         1609.2 PSID (Protocol Service Identifier),specified as an integer.
+         *   - channel:      The channel on which the message should be  transmitted,specified as an integer. 
+         */
+        class RawSpdu : public tmx::message
+        {
+        public:
+            RawSpdu() {}
+            RawSpdu(const tmx::message_container_type &contents) : tmx::message(contents) {}
+            // Full SPDU bytes
+            std_attribute(this->msg, tmx::byte_stream, spdu_data, tmx::byte_stream(), )
```
adev4a: We may just want to pick snake/camelCase everywhere here for all variables defined here

adev4a (changes requested):

src/tmx/Messages/include/RawSpdu.h
mpalash95: renamed all variables to camelCase

src/tmx/Messages/test/RawSpduTest.cpp
```diff
@@ -0,0 +1,100 @@
...
+    msg.setSpduDataBytes(std::make_shared<std::vector<uint8_t>>(spdu));
+    msg.set_uuid(uuid);
+    msg.set_messageType("BSM");
+    msg.set_timestampMs(1718900000000);
+    msg.set_psid(8002);
+    msg.set_channel(183);
+
+    std::string json = msg.to_string();
+
+    EXPECT_NE(json.find("\"spduData\":\"" + bsmHex + "\""), std::string::npos);
```
paulbourelly999, line 65: NIT: I don't think this is what you meant to do. I think EXPECT_NE does a comparison to see if two pointers **DONT** point to the same memory location. I think this works because json.find returns a size_t value of where some data is found which is of course not equal to std::string:npos unless the value is not found. Functionally this works although I think it is far easier to just to a value comparison between the entire returned string and some expected value.

src/tmx/Messages/test/RawSpduTest.cpp
```diff
@@ -0,0 +1,100 @@
...
+    EXPECT_NE(json.find("\"channel\":\"183\""), std::string::npos);
+}
+
+TEST(RawSpduTest, SetAndGetAttributesPsm) {
+    RawSpdu msg;
+
+    tmx::byte_stream spdu = tmx::byte_stream_decode(psmHex);
+    tmx::byte_stream uuid = tmx::byte_stream_decode(uuidHex);
+
+    msg.setSpduDataBytes(std::make_shared<std::vector<uint8_t>>(spdu));
```
paulbourelly999: Why does this need to be convert to uint8_t and the UID does not. This should be storable as the byte stream in the message

src/tmx/Messages/include/RawSpdu.h
```diff
@@ -0,0 +1,66 @@
...
+                /// 1609Dot2 PSID.
+                std_attribute(this->msg, int, psid, 0, )
+                ///  channel Number
+                std_attribute(this->msg, int, channel, 0, )
+                // Set the SPDU data as a vector of bytes
+                void setSpduDataBytes(std::shared_ptr<std::vector<uint8_t>> bytes)
+            {
+                set_spduData(tmx::byte_stream(bytes->begin(), bytes->end()));
+            }
+            std::shared_ptr<std::vector<uint8_t>> getSpduDataBytes()
```
paulbourelly999: This does not need to be a pointer. Also this should return byte_stream not vector<uint8_t>

src/tmx/Messages/include/RawSpdu.h
```diff
@@ -0,0 +1,66 @@
...
+                // J2735 payload type, e.g. "BSM", "PSM", "MAP".
+                std_attribute(this->msg, std::string, messageType, "", )
+                // timestamp
+                std_attribute(this->msg, int64_t, timestampMs, 0, )
+                /// 1609Dot2 PSID.
+                std_attribute(this->msg, int, psid, 0, )
+                ///  channel Number
+                std_attribute(this->msg, int, channel, 0, )
+                // Set the SPDU data as a vector of bytes
+                void setSpduDataBytes(std::shared_ptr<std::vector<uint8_t>> bytes)
```
paulbourelly999: same comment as above

src/tmx/Messages/include/RawSpdu.h
```diff
@@ -0,0 +1,66 @@
...
+                std_attribute(this->msg, tmx::byte_stream, uuid, tmx::byte_stream(), )
+
+                // J2735 payload type, e.g. "BSM", "PSM", "MAP".
+                std_attribute(this->msg, std::string, messageType, "", )
+                // timestamp
+                std_attribute(this->msg, int64_t, timestampMs, 0, )
+                /// 1609Dot2 PSID.
+                std_attribute(this->msg, int, psid, 0, )
+                ///  channel Number
+                std_attribute(this->msg, int, channel, 0, )
```
paulbourelly999: @kjrush I am not sure this information means anything to us. Channel is not included in the 1609.2 header. It is an RSU level configuration for each PSID message type. It should be pretty meaningless between the two proxy radios unless I am mistaken

paulbourelly999: In other words why would proxy radio A care what channel proxy radio B broadcasts a given individual message? This would be a RSU level PSID configuration on both RSUs and would be lost as soon as we exit the proxy radios/DT network and replaced by what ever channel the device under test received or broadcast the message under.

paulbourelly999 (changes requested):

src/tmx/Messages/test/RawSpduTest.cpp
mpalash95: Removed the shared pointer and assigned tmx::byte_stream to spdu

src/tmx/Messages/include/RawSpdu.h
mpalash95: changed

src/tmx/Messages/include/RawSpdu.h
mpalash95: Addressed

src/tmx/Messages/include/RawSpdu.h
kjrush: Mostly I was looking to match the existing legacy data patterns in the existing DSRCMetadata struct, I'd assumed that meant we were getting the channel information from the RSU directly. 

Given that we only have a single channel for V2X now, it's not particularly relevant, but in a world where multiple channels were available it would be important that the channel a device-under-test broadcasted a message on should be preserved through the HWIL system. (e.g. DUT1 broadcasts a BSM on 182, the proxy radio for DUT2 should likewise broadcast that message on 182).

If it's no longer relevant and/or not available I see no reason to have these fields.

mpalash95: Removed channel number from RawSPDU

paulbourelly999 (dismissed):

adev4a (approved):

PR #908, opened 2026-06-22
Fix security issue conditional mutex lock

paulbourelly999:
- prevent unintended deadlock

# PR Details
## Description
Fix security/maintainability issue reported with conditional mutex lock
[Issue](https://sonarcloud.io/project/issues?impactSoftwareQualities=SECURITY&issueStatuses=OPEN%2CCONFIRMED&id=usdot-fhwa-ops_V2X-Hub&open=AZ7vkj2GWPN9veXDLph1)

## Related GitHub Issue
NA

## Related Jira Key
NA

## Motivation and Context
Address security/maintainability issue

## How Has This Been Tested?
CI Integration Tests and local integration testing of Message Receiver and FLIR Camera Driver plugins with configuration changes

## Types of changes

## Checklist:


dan-du-car:
Would like to confirm that the solution to this issue is to remove the lock to prevent deadlock? Is lock no longer required for the config update?

paulbourelly999:
> ve the lock to prevent deadlock? Is lock no longer required for the config update?

The lock is now built into the PluginClient (see https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/908/changes#diff-06f23582e5ab569d0cab42ce69d5df666e4f935ff706b0e02a6b0841da50b27aR385). This reduces duplication, where each plugin would have its out configuration mutex locking structure and removes the possibility of thread dead lock with conditional lock statements as well

src/tmx/TmxUtils/src/PluginClient.h
```diff
@@ -213,8 +212,7 @@ class PluginClient: public Runnable {
 			free(text);
 		}
 
-		if (lock != nullptr)
-			lock->unlock();
+		_configLock.unlock();
```
dan-du-car: Nit: Does it make sense to use unique lock or lock guard in case unknown exception is thrown and is not caught?

dan-du-car (dismissed):

src/tmx/TmxUtils/src/PluginClient.h
paulbourelly999: I think so. I will change this

dan-du-car (approved):

PR #910, opened 2026-06-24
add spdu processing

adev4a:
# PR Details
This PR adds SPDU processing to the MessageReceiver Plugin. An optional flag for `FullSPDUMode` is added to the plugin which updates the processing for the incoming message in the plugin. When set, the incoming messages are handled as 1609.2 SPDUs where the message payload is preserved as a RawSPDU tmx message.

The message is decoded as a COER as per definition in : https://github.com/usdot-fhwa-stol/stol-1609dot2/blob/bc7149771b08d1fc1e8afab37b840b3554c81334/1609dot2-2022/asn/Ieee1609Dot2.asn#L138

Supporting Integration testing PR: https://github.com/usdot-fhwa-OPS/v2xhub-integration-tests/pull/18

image

This PR requires CI enabled Integration test updates to test for these changes, implemented under: https://github.com/usdot-fhwa-OPS/v2xhub-integration-tests/pull/17

## Related Jira Key

https://usdot-carma.atlassian.net/browse/DT4IT-104
## How Has This Been Tested?

Locally Integration Tested

## Types of changes

## Checklist:


src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
```diff
@@ -125,6 +126,100 @@ void MessageReceiverPlugin::OnStateChange(IvpPluginState state)
 	}
 }
 
+bool unwrapSpdu(const Ieee1609Dot2Data_t* d, std::vector<uint8_t>& payloadOut, uint32_t& psidOut, bool& psidSet)
```
kjrush: I like the overall recursive design here.

Looks like this is just a private helper method, which is good, but might not be a bad idea to write a short doc comment here.  I think there's a couple of rough edges here that might be worth noting for future readers/users:

1. psidOut and psidSet are not overwritten unless a PSID is actually found within the message. If for some reason a PSID is not found in the 16092Content (for example, it just contains unsecured data) then the value of psidSet will remain whatever it was before this function call (even if that was `true`, despite no PSID being found), same with psidOut. 
2. psidOut and psidSet might be modified even if the overall return is false and the unwrap operation failed.

Both of these could potentially be fixed by kind of bootstrapping this recursion by defining a second function that sets the initial state for the recursion and then checks the result of the recursion before overwriting the caller's output variables via the references.

e.g.

```
bool unwrapSpdu(const Ieee1609Dot2Data_t* d, std::vector<uint8_t>& payloadOut, uint32_t& psidOut, bool& psidSet) {
    bool tmpPsidSet = false;
    uint32_t tmpPsidOut = 0;
    // Could potentially even isolate the payloadOut vector, but seems unnecessary since the return value already indicates whether it was successfully modified (and that would potentially memory inefficient)
    bool success = unwrapSpduHelper(d, payloadOut, tmpPsidSet, tmpPsidOut);
    if (success && tmpPsidSet && tmpPsidOut != 0) {
        psidOut = tmpPsidOut;
        psidSet = tmpPsidSet;
    }
   return success;
}
```

Where you'd rename the current `unwrapSpdu()` to `unwrapSpduHelper()` for clarity that this outer function is the one intended for usage. You could alternatively refactor unwrapSpduHelper to be a recursive lambda function defined within `unwrapSpdu()` but that's sacrificing clarity in other ways to gain clarity that that helper function isn't supposed to be called outside of the main `unwrapSpdu()`, probably not a good tradeoff

To be honest, I'm not entirely sure it's worth refactoring like this given that its just a helper function, but wanted to point out some potential rough edges here and how they could be addressed, which would be desirable if this were a function that might be called by some external user of this class.

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
```diff
@@ -125,6 +126,100 @@ void MessageReceiverPlugin::OnStateChange(IvpPluginState state)
...
+            if (!next) return false;
+            return unwrapSpdu(next, payloadOut, psidOut, psidSet);
+        }
+
+        default:
+            return false;                // unknown CHOICE
+    }
+}
+
+long MessageReceiverPlugin::identifyJ2735Type(const std::vector<uint8_t>& payload)
```
kjrush: Is there a reason for using long here over int64_t? Generally I think it's better practice to explicit about integer sizes rather than relying on long/short/etc. as those can vary based on processor architecture.

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
```diff
@@ -125,6 +126,100 @@ void MessageReceiverPlugin::OnStateChange(IvpPluginState state)
...
+{
+    std::stringstream ss;
```
kjrush: Is there a reason we're doing this as strings? It's not the end of the world, but I think using strings for non-textual data is a code smell generally. At minimum a hex-string version of a byte-array takes up twice or more as much memory, and marshalling back and forth takes computation time. Unless this is being presented to the end user, or we're integrating with a legacy system which has already has the issue of working with strings, generally I'd rather we work with vectors or arrays of `uint8_t`s

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
```diff
@@ -187,23 +294,23 @@ int MessageReceiverPlugin::Main()
...
 
-					string cmd1="curl -X POST "+url+" -H \'Content-Type: application/json\' -d "+req; 
+					string cmd1="curl -X POST "+url+" -H \'Content-Type: application/json\' -d "+req;
 
-					const char *cmd=cmd1.c_str();  
+					const char *cmd=cmd1.c_str();
 					char buffer[2048];
 					std::string result="";
-					FILE* pipemsg= popen(cmd,"r"); 
+					FILE* pipemsg= popen(cmd,"r");
```
kjrush, line 220: I feel like I'm missing some significant context here. Why are we invoking curl and creating new file descriptors during the act of processing a received message?

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
```diff
@@ -125,6 +126,100 @@ void MessageReceiverPlugin::OnStateChange(IvpPluginState state)
...
+    ss << std::hex << std::setfill('0');
+    size_t n = std::min<size_t>(payload.size(), IDCHECKLIMIT / 2);
+    for (size_t i = 0; i < n; ++i)
+        ss << std::setw(2) << static_cast<unsigned>(payload[i]);
+
+    size_t idloc; int mlen; long msgId;
+    return findMessageId(ss.str(), idloc, mlen, msgId) ? msgId : -1;
+}
+
+bool MessageReceiverPlugin::findMessageId(const std::string& hex, size_t& idloc,
```
kjrush: This probably works, but I don't quite like it. We're value-hunting for a known message ID somewhere in a blob, even though we know that blob is structured according to some schema. It's also not guaranteed that the ID value is byte-aligned (the U in UPER) and this approach only 100% works on byte-aligned data anyway. Functionally, I can't think of a case in which that would fail since MsgID is very early on in the MsgFrame object so it's almost certainly byte-aligned, but there's something fundamentally about making an assumption like that rather than simply properly parsing it that is raising flags for me.

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
```diff
@@ -125,6 +126,100 @@ void MessageReceiverPlugin::OnStateChange(IvpPluginState state)
...
+            idloc    = loc;
+            hexLen   = mlen;
+            dsrcMsgId = strtol(id.c_str(), nullptr, 16);
+            return true;
+        }
+    }
+    return false;
+}
+
+bool MessageReceiverPlugin::buildSecuredV2XMessage(const byte_stream& incoming, int len, uint64_t rxTime, SecuredV2XMessage& out, std::vector<uint8_t>& payloadOut)
```
kjrush: It looks like this largely functions like my proposed wrapper function from earlier for unwrapSpdu, so might not be as relevant. Still think it wouldn't hurt to document those edge cases I mentioned as a comment, but the larger refactor doesn't seem merited based on this.

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
```diff
@@ -125,6 +126,100 @@ void MessageReceiverPlugin::OnStateChange(IvpPluginState state)
...
+{
+    // 1. Decode the SPDU
+    Ieee1609Dot2Data_t* decodedPtr = nullptr;
+	// Add decode message logic
+
+    // 2. Walk to the inner J2735 payload
+    uint32_t psid = 0;
+	bool psidSet = false;
```
kjrush: This value is defined but not used, other than being passed in (as an output parameter) to unwrapSpdu. Is it not necessary?

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
```diff
@@ -125,6 +126,100 @@ void MessageReceiverPlugin::OnStateChange(IvpPluginState state)
...
+    }
+
+    // 3. Populate the struct
+    out.spdu = std::make_shared<const std::vector<uint8_t>>(
+                            incoming.data(), incoming.data() + len);
+    out.id            = _uuidGen();
+    out.rxTimestampMs = rxTime;
+    out.dsrc.psid     = psid;
+    out.dsrc.channel  = _dsrcChannel;
+    out.j2735Type = identifyJ2735Type(payloadOut);
```
kjrush: Would we be able to inject SPDUs of messages V2Xhub is unable to understand? Lets say somebody was using a message from a J2735 add-on that we hadn't added support to V2Xhub yet, would we still be able to propagate that through for Radio HWIL? Or does V2Xhub need to actually have definitions for every message we pass through?

Just asking for awareness sake.

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
```diff
@@ -159,15 +254,27 @@ int MessageReceiverPlugin::Main()
 				uint64_t time = Clock::GetMillisecondsSinceEpoch();
 
 				totalBytes += len;
-				int txlen=0; 
-				
+				int txlen=0;
+
 				// @SONAR_STOP@
 				// if verification enabled, access HSM
+				if (fullSPDUMode == true){
```
kjrush: nit, `if (bool == true)` is redundant compared to `if (bool)`

kjrush (commented):
Mostly minor stuff, I'll leave to your judgment if you want to make any adjustments. Functionally this seems fine, I have some opinions on data types and message type identification but they're not hills I'm willing to die on if the alternatives represent substantially more work.

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
adev4a: An artifact of the original code moved to a function here, updated to int64_t

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
adev4a: Also an artifact of the original code. Will double check with @paulbourelly999 if this is a certain way for a reason

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
adev4a: Tagging @paulbourelly999 again, hesitant to update some of this without confirming this wasn't implemented a certain reason. Will update with comments if it was.

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
adev4a: Also refactored into an independent function based on existing implementation, tagging @paulbourelly999, for discussion.

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
adev4a: Maybe i'm misunderstanding, or this was fixed after this comment but psidSet is being used to check whether the psid is set in the recursion call unwrapSpdu.

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
adev4a: Updated

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
adev4a: This is a good point. I don't see any reason we shouldn't be able to still keep the SPDUs of all messages if publish the raw spdu to the tmx core before we start processing the J2735 message. Will add this update.

adev4a:
Github actions were temporarily updated and integration tested successfully against https://github.com/usdot-fhwa-OPS/v2xhub-integration-tests/pull/17

image

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
paulbourelly999: This was from an old implementation where V2X Hub was verifying message signatures using an HSM by interacting with a REST API hosted by the HSM device. This logic is only active when enabling this functionality and will likely soon be removed.

paulbourelly999: It looks like your changes here are just formatting. I would recommend disabling adding autoformatting you have setup on file save to avoid the PR including line changes that are not intentional

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
```diff
@@ -125,6 +127,114 @@ void MessageReceiverPlugin::OnStateChange(IvpPluginState state)
 	}
 }
 
+bool MessageReceiverPlugin::unwrapSpdu(const Ieee1609Dot2Data_t* d, std::vector<uint8_t>& payloadOut, uint32_t& psidOut, bool& psidSet)
```
paulbourelly999: This function does not need access to Plugin state. Make this a free function so you can unit test it

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
```diff
@@ -125,6 +127,114 @@ void MessageReceiverPlugin::OnStateChange(IvpPluginState state)
...
+			}
+            return unwrapSpdu(next, payloadOut, psidOut, psidSet);
+        }
+
+        default:
+            return false;                // unknown CHOICE
+    }
+}
+
+int64_t MessageReceiverPlugin::identifyJ2735Type(const std::vector<uint8_t>& payload)
```
paulbourelly999: same comment as above

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
paulbourelly999: It looks to me like this message ID search is attempting to parse out the full SPDU header and find where the payload starts. With you new changes we should no longer need this right? We be able to read the SPDU header exclicitlty. I think we just have to account for the case where no header is sent, but in that case I do not think we need to find the message ID either.

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.cpp
```diff
@@ -125,6 +127,114 @@ void MessageReceiverPlugin::OnStateChange(IvpPluginState state)
...
+            idloc    = loc;
+            hexLen   = mlen;
+            dsrcMsgId = strtol(id.c_str(), nullptr, 16);
+            return true;
+        }
+    }
+    return false;
+}
+
+bool MessageReceiverPlugin::buildRawSpduMessage(const byte_stream& incoming, int len, uint64_t rxTime, tmx::messages::RawSpdu& out, std::vector<uint8_t>& payloadOut)
```
paulbourelly999: This can be a free function as well so we can unit test it

src/v2i-hub/MessageReceiverPlugin/src/MessageReceiverPlugin.h
```diff
@@ -62,18 +67,25 @@ class MessageReceiverPlugin: public tmx::utils::TmxMessageManager {
...
 	tmx::utils::FrequencyThrottle<int> errThrottle;
 	tmx::utils::FrequencyThrottle<int> statThrottle;
 	uint _skippedSignVerifyErrorResponse;
+	uint _processedSPDU = 0;
 	const char* Key_SkippedSignVerifyError = "Message Skipped (Signature Verification Error Response)";
+	const char* Key_ProcessedSPDU = "Message Processed (SPDU)";
+	long identifyJ2735Type(const std::vector<uint8_t>& payload);
+	bool findMessageId(const std::string& hex, size_t& idloc, int& hexLen, long& dsrcMsgId);
+	bool buildRawSpduMessage(const byte_stream& incoming, int len, uint64_t rxTime, tmx::messages::RawSpdu& out, std::vector<uint8_t>& payloadOut);
+	bool unwrapSpdu(const Ieee1609Dot2Data_t* d, std::vector<uint8_t>& payloadOut, uint32_t& psidOut, bool& psidSet);
```
paulbourelly999: Please write javadoc style for any methods or functions your create

src/v2i-hub/MessageReceiverPlugin/CMakeLists.txt
```diff
@@ -7,9 +7,10 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 SET (TMX_PLUGIN_NAME "MessageReceiver")
 SET (CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/modules)
 
-
+set(1609dot2_LIBRARY_NAME "::stol-1609dot2-2022")
```
paulbourelly999: NIT: What is the point of this set call. We only link this library once here so the variable really only makes this less concise

paulbourelly999 (changes requested):

adev4a:
Closed by https://github.com/usdot-fhwa-OPS/V2X-Hub/pull/922

PR #919, opened 2026-07-09
Fix default IP in initialization script

paulbourelly999:
# PR Details
## Description
Due to the network security improvements added to V2X Hub outlined here (https://github.com/usdot-fhwa-OPS/V2X-Hub/blob/develop/configuration/NETWORK_SECURITY.md#migration-considerations) loopback of V2X-Hub container no longer resolves to host system loop back. This means when configuring V2X Hub IP you want to use 0.0.0.0 to configure listening actors inside the V2X Hub container to receive messages sent to the HOST machine.

Additionally added prompts to configure V2XHUB_VOLUME_PATH.

Also added fix for INFRASTRUCTURE_NAME

## Related GitHub Issue
NA

## Related Jira Key
NA

## Motivation and Context
Fix initialization script

## How Has This Been Tested?
NA

## Types of changes

## Checklist:

### DIFF
--- a/configuration/initialize_docker_environment.sh
+++ b/configuration/initialize_docker_environment.sh
@@ -42,9 +42,8 @@ if [[ "$interactive" == true ]]; then
     PORT_DRAYAGE_ENABLED_DEFAULT="FALSE"
     INFRASTRUCTURE_ID_DEFAULT="rsu_1234"
     INFRASTRUCTURE_NAME_DEFAULT="East Intersection"
-    V2XHUB_IP_DEFAULT="127.0.0.1"
+    V2XHUB_IP_DEFAULT="0.0.0.0"
     SIMULATION_MODE_DEFAULT="FALSE"
-    SIMULATION_IP_DEFAULT="127.0.0.1"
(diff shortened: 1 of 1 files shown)


adev4a:
These changes look good. The initialization script is currently also overwriting the V2XHub Volume path - since that is not requested from the user and only the user input is populated in the .env. This is also breaking deployment using the script. Can we add a user prompt for the volume as well or alternatively include a default value?

configuration/initialize_docker_environment.sh
```diff
@@ -42,10 +42,10 @@ if [[ "$interactive" == true ]]; then
     PORT_DRAYAGE_ENABLED_DEFAULT="FALSE"
     INFRASTRUCTURE_ID_DEFAULT="rsu_1234"
     INFRASTRUCTURE_NAME_DEFAULT="East Intersection"
```
adev4a, line 44: The included space in the default Intersection Name here is also breaking this script. Can we fix the parsing or update the default to East_Intersection?

paulbourelly999: I wonder why this breaks stuff. I will confirm whether this breaks anything and why and attempt to fix this

adev4a (approved):

PR #927, opened 2026-07-27
Fix heap after use fault

paulbourelly999:
# PR Details
## Description
This PR attempts to make the final fixes for memory management for TmxJ2735Message and TmxJ2735EncodedMessage. These are intermediate classes used to handle J2735 messages. This includes:

-  Serializing/Deserializing messages to and from UPER/JSON/XML
-  Message handling for plugins to broadcast or subscribe to messages

These classes are wrapper types on the C structs defined in stol-j2735 library, generated from the ASN1 schema definitions in SAE J2735.

> [!CAUTION]
> This change impacts lifetime of core V2X message classes used in almost all plugins and use cases. Due to the changes introduced in this PR, this could introduce new failures like double free segfaults for plugins that had implemented work arounds. This PR requires substantial testing to ensure functionality is not regressed.

### Bug Description 
In V2X Hub 7.11.1 there is a bug that using `AddMessageFilter<TmxJ2735EncodeMessage<T>>()` would cause a **heap-use-after-free** segmentation fault when trying to use the returned `&TmxJ2735EncodedMessage<T>` reference in the handler call back because it was being destroyed.  Compiling in debug mode would result in a Asan error looking something like this 
```
=================================================================
==423848==ERROR: AddressSanitizer: heap-use-after-free on address 0x516000027ca0 at pc 0x60c27c3704da bp 0x709e247ef640 sp 0x709e247ef630
READ of size 8 at 0x516000027ca0 thread T2
    #0 0x60c27c3704d9 in CARMACloudPlugin::CARMACloudPlugin::HandleCARMARequest(tmx::messages::TmxJ2735Message<TestMessage04, battelle::attributes::XML>&, tmx::tmx_routeable_message<battelle::attributes::JSON>&) /workspace/src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp:50
    #1 0x60c27c4fc644 in tmx::utils::PluginClient::handler_allocator_impl<tmx::messages::TmxJ2735Message<TestMessage04, battelle::attributes::XML>, CARMACloudPlugin::CARMACloudPlugin, CARMACloudPlugin::CARMACloudPlugin>::invokeHandler(tmx::tmx_routeable_message<battelle::attributes::JSON>&) (/var/www/plugins/CARMACloudPlugin/bin/CARMACloudPlugin+0x1d7644)
    #2 0x60c27c64e089 in tmx::utils::PluginClient::invoke_handler(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, tmx::tmx_routeable_message<battelle::attributes::JSON>&) /workspace/src/tmx/TmxUtils/src/PluginClient.cpp:444
    #3 0x60c27c6508f6 in tmx::utils::PluginClient::OnMessageReceived(IvpMessage*) /workspace/src/tmx/TmxUtils/src/PluginClient.cpp:330
    ...
0x516000027ca0 is located 32 bytes inside of 632-byte region [0x516000027c80,0x516000027ef8)
freed by thread T2 here:
    #0 0x709e2f5b6537 in __interceptor_free ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:127
    #1 0x60c27c362a85 in j2735_destroy<tmx::messages::j2735::SaeJ2735Traits<MessageFrame> > /workspace/src/tmx/TmxApi/tmx/messages/SaeJ2735Traits.hpp:106

previously allocated by thread T2 here:
    #0 0x709e2f5b6a57 in __interceptor_calloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:154
    #1 0x709e2dcbb73d in SEQUENCE_decode_uper (/opt/carma/lib/libstol-j2735-2024-carma.so.1+0xf573d)

Thread T2 created by T0 here:
    #0 0x709e2f55a685 in __interceptor_pthread_create ../../../../src/libsanitizer/asan/asan_interceptors.cpp:216
    #1 0x709e2ddb01f8 in ivp_create /workspace/src/tmx/TmxApi/tmx/IvpPlugin.c:79
    #2 0x7ffd8b88803f  ([stack]+0x1f03f)

SUMMARY: AddressSanitizer: heap-use-after-free /workspace/src/v2i-hub/CARMACloudPlugin/src/CARMACloudPlugin.cpp:50 in CARMACloudPlugin::CARMACloudPlugin::HandleCARMARequest(tmx::messages::TmxJ2735Message<TestMessage04, battelle::attributes::XML>&, tmx::tmx_routeable_message<battelle::attributes::JSON>&)
Shadow bytes around the buggy address:
 ...
```
This was ultimately because the TmxJ2735EncodedMessage<T> and its pointer to TmxJ2735Message<T> would go out of scope and the destructor would be called to clean up the underlying memory of the T pointer. 

The reason the `TmxJ2735EncodedMessage<T>` and its pointer would go out of scope is due to the following code snippet in the `PluginClient.cpp` message handling.
```cpp
template <typename MsgType, class PluginType, class HandlerType>
struct handler_allocator_impl: public handler_allocator {
      using msg_type = MsgType;
      
      handler_allocator_impl(PluginType *plugin,
		      void (HandlerType::*handler)(MsgType &, tmx::routeable_message &)):
			      instance(plugin), fn(handler) {}
      
      std::string get_messageType() override
      {
	      return battelle::attributes::type_id_name<MsgType>();
      }
      
      void invokeHandler(tmx::routeable_message &routeableMsg) override
      {
              // For AddFilter calls for J2735 Messages MsgType == TmxJ2735EncodedMessage<T> where 
              // T is the ASNC J2735 message struct like SPAT. The method get_payload<MsgType> is used to 
              // convert a generic routable_message to the specific type of message you want. When called
              // with T as TmxJ2735EncodedMessage, it creates a temporary TmxJ2735EncodedMessage using the 
              //  hex bytes stored in the routeable_message and then calls decode_j2735_message() which ultimately
              // returns a lref to the TmxJ2735EncodedMessage._decoded C struct stored in the private shared pointer
              //. Since the parent object goes out of scope immediately, since nothing has a reference to it our its interal 
              // shared pointer, its left dangling. 
	      MsgType msg = routeableMsg.template get_payload<MsgType>();
	      if (fn)
		      (instance->*fn)(msg, routeableMsg);
	      else
		      throw PluginException("Missing handler for " + get_messageType());
      }
private:
      PluginType *instance;
      void (HandlerType::*fn)(MsgType &, tmx::routeable_message &);
};
```
To fix this, we wanted to add the internally managed shared pointer as a local variable to ensure it does not go out of scope before the handler call completes. This was implemented using the `J2735MessageFactory` to convert the `routeable message` into a `TmxJ2735EncodedMessage<T>` and using SFINAE (Substitution Failure is Not An Error) macros to conditionally trigger this only for `MsgType` == `TmxJ2735EncodedMessage<T>` and not for other broadcasted `routeable_msg`.  The new Plugin Client Logic is below:
```cpp
void invokeHandler(tmx::routeable_message &routeableMsg) override
{
	MsgType msg = routeableMsg.template get_payload<MsgType>();
	if (fn)
		(instance->*fn)(msg, routeableMsg);
	else
		throw PluginException("Missing handler for " + get_messageType());
	if ( instance->IsJ2735Message(routeableMsg)) {
		
		if constexpr (tmx::messages::j2735::is_instance_of_v<MsgType, tmx::messages::TmxJ2735Message>) {
			tmx::messages::J2735MessageFactory factory;
			tmx::byte_stream bytes = routeableMsg.get_payload_bytes();
			std::shared_ptr<tmx::messages::TmxJ2735EncodedMessage<MsgType>> encodeMsg(
				static_cast< tmx::messages::TmxJ2735EncodedMessage<MsgType> * > (factory.NewMessage(bytes))
			);
			MsgType msg = encodeMsg->decode_j2735_message();
			if (fn)
				(instance->*fn)(msg, routeableMsg);
			else
				throw PluginException("Missing handler for " + get_messageType());
		}
		else {
			throw PluginException("Missing J2735 handler for " + get_messageType());
		}
	}
	else {	
		MsgType msg = routeableMsg.template get_payload<MsgType>();
		if (fn)
			(instance->*fn)(msg, routeableMsg);
		else
			throw PluginException("Missing handler for " + get_messageType());
	}
}

```

And the new SFINAE logic to check if `MsgType` template is an instance of `TmxJ2735EncodeMessage` which it self is a templated type is here 

```cpp
// Primary template: returns false for everything else
template <typename T, template <typename...> class Template>
struct is_instance_of : std::false_type {};

// Partial specialization: matches when T is an instantiation of Template
template <template <typename...> class Template, typename... Args>
struct is_instance_of<Template<Args...>, Template> : std::true_type {};

/**
 * Used to determine whether type is instance of template class.
 * std::is_same_v only works with full defined classes.
 */
template <typename T, template <typename...> class Template>
inline constexpr bool is_instance_of_v = is_instance_of<T, Template>::value;

```

## Related Jira Key
[VH-1564
](https://usdot-carma.atlassian.net/browse/VH-1564)

## Motivation and Context
Fix memory management for V2X Hub.

## Types of changes

## Checklist:


src/tmx/TmxApi/tmx/messages/TmxJ2735.hpp
```diff
@@ -115,7 +115,7 @@ class TmxJ2735Message: public tmx::tmx_message<Format>
 	 * Copy from existing shared pointer of same type.  Current ownership is still
 	 * maintained in the existing shared pointer, but reference count is increased.
 	 */
-	TmxJ2735Message(const std::shared_ptr<message_type> &other):
+	TmxJ2735Message(const std::shared_ptr<message_type> other):
```
paulbourelly999, line 118: Comment says it copies shared pointer but using const shared_ptr & does not copy the shared pointer and does not increase reference count.

src/tmx/TmxApi/tmx/messages/TmxJ2735.hpp
```diff
@@ -125,8 +125,21 @@ class TmxJ2735Message: public tmx::tmx_message<Format>
 	 */
 	template <typename OtherMsgType>
 	TmxJ2735Message(const std::shared_ptr<OtherMsgType> &other):
-		tmx::tmx_message<Format>(),
-		_j2735_data(j2735::j2735_cast<message_type>(other.get()), [](message_type *p) { }) { }
+		tmx::tmx_message<Format>()
+		,_j2735_data(j2735::j2735_cast<message_type>(other.get()), [](message_type *p) {
```
paulbourelly999, line 129: SaeJ2735Traits.hpp has static functions to cast down and up message types like SPAT into the full MessageFrame. This was implemented to allow for users to be able to get only the V2X decoded message they want from the original UPER bytes but still send and be able to encode the complete message frame for receiving and forwarding. This constructor extends this capability to the TmxJ2735Message but memory management of the underlying pointer becomes more difficult with this casting since both the TmxJ2735Message<MessageFrame> and the TmxJ2735Message<SPAT> will include a shared pointer to the *SPAT memory. During cleanup when the TmxJ2725Message<MessageFrame> only owns the C struct for MessageFrame while the TmxJ2735Message<SPAT> will use the ASN_STRUCT_FREE recursive call to free SPAT and all the internal pointers.

src/tmx/TmxUtils/src/kafka/kafka_consumer_worker.cpp
```diff
@@ -124,8 +124,7 @@ namespace tmx::utils
 
     const char *kafka_consumer_worker::consume(int timeout_ms)
     {
-        const RdKafka::Message *msg = nullptr;
-        msg = _consumer->consume(timeout_ms);
+        std::shared_ptr<RdKafka::Message> msg(_consumer->consume(timeout_ms));
```
paulbourelly999, line 127: Fixes memory leak in kafka client implementation

src/tmx/TmxUtils/src/PluginClient.h
```diff
@@ -405,11 +408,33 @@ class PluginClient: public Runnable {
...
+					tmx::messages::J2735MessageFactory factory;
+					tmx::byte_stream bytes = routeableMsg.get_payload_bytes();
+					std::shared_ptr<tmx::messages::TmxJ2735EncodedMessage<MsgType>> encodeMsg(
+						static_cast< tmx::messages::TmxJ2735EncodedMessage<MsgType> * > (factory.NewMessage(bytes))
+					);
+					MsgType msg = encodeMsg->decode_j2735_message();
+					if (fn)
+						(instance->*fn)(msg, routeableMsg);
+					else
+						throw PluginException("Missing handler for " + get_messageType());
```
MishkaMN, line 428: So this `invokeHandler` function will not be possible to be called with something like `MsgType = TmxJ2735EncodedMessage<SpatMessage>` right? when the msg type is not `TmxJ2735Message` but `TmxJ2735EncodedMessage`. In which case, this will keep returning false as they are not equal

If it is possible to be called like that, maybe we should add another if clause like:
```
else if constexpr (tmx::messages::j2735::is_instance_of_v<MsgType, tmx::messages::TmxJ2735EncodedMessage>) {
    // MsgType IS the encoded wrapper already
    MsgType msg(routeableMsg);
    if (fn)
        (instance->*fn)(msg, routeableMsg);
    else
        throw PluginException("Missing handler for " + get_messageType());
}
```

src/tmx/TmxUtils/src/PluginClient.h
```diff
@@ -405,11 +408,33 @@ class PluginClient: public Runnable {
...
-				(instance->*fn)(msg, routeableMsg);
-			else
-				throw PluginException("Missing handler for " + get_messageType());
+			if ( instance->IsJ2735Message(routeableMsg)) {
+				
+				if constexpr (tmx::messages::j2735::is_instance_of_v<MsgType, tmx::messages::TmxJ2735Message>) {
+					tmx::messages::J2735MessageFactory factory;
+					tmx::byte_stream bytes = routeableMsg.get_payload_bytes();
+					std::shared_ptr<tmx::messages::TmxJ2735EncodedMessage<MsgType>> encodeMsg(
+						static_cast< tmx::messages::TmxJ2735EncodedMessage<MsgType> * > (factory.NewMessage(bytes))
```
MishkaMN: Looks like this NewMessage function can return NULL. Will that be issue trying to directly use it to decode?

MishkaMN: I am assuming we don't want to crash (it might segfault? if it was run on NULL?) if any invalid msg comes and just toss it out. Then, maybe adding a guard here like:
```
tmx::messages::J2735MessageFactory factory;
tmx::byte_stream bytes = routeableMsg.get_payload_bytes();

auto *raw = factory.NewMessage(bytes);
auto *typed = dynamic_cast<tmx::messages::TmxJ2735EncodedMessage<MsgType> *>(raw);

if (!typed) {
    delete raw;  // factory allocated *something* of the wrong type, don't leak it
    throw PluginException("Unable to decode J2735 message for " + get_messageType() +
                           (raw ? ": message type mismatch" : ": " + factory.get_event().what()));
}

std::shared_ptr<tmx::messages::TmxJ2735EncodedMessage<MsgType>> encodeMsg(typed);
MsgType msg = encodeMsg->decode_j2735_message();
```

src/tmx/TmxUtils/src/PluginClient.h
MishkaMN: Not blocking though, I won't be back in time, just something think about

src/tmx/TmxApi/tmx/messages/SaeJ2735Traits.hpp
```diff
@@ -104,10 +123,63 @@ static void j2735_destroy(typename TraitsType::message_type *ptr,
 	{
 		typename TraitsType::asn_type *asnType = (typename TraitsType::asn_type *)descr;
 		ASN_STRUCT_FREE((*asnType), ptr);
+		ptr = nullptr;
```
liuji1031: line 126 and 130: setting ptr to nullptr does not have real impact since it is passed in by value. Do we need a reference to the pointer instead if the intention is setting the pointer to null after free?

src/tmx/TmxUtils/src/PluginClient.h
paulbourelly999: I do not understand. The expression checks if MsgType is an instance of TmxJ2735EncodedMessage. This is the message type we use to broadcast J2735 Messages. What other type are we concerned with this returning false for

src/tmx/TmxApi/tmx/messages/SaeJ2735Traits.hpp
paulbourelly999: This is not pass by value. Check function signature `(typename TraitsType::message_type *ptr`, It passes the pointer directly. Am I missing something here.

src/tmx/TmxUtils/src/PluginClient.h
paulbourelly999: Resolved. No point to free what factory returns since it is null ptr. Also the exception information about the msg type and bytes.

liuji1031 (dismissed):
The changes look good to me.

src/tmx/TmxApi/tmx/messages/TmxJ2735Codec.hpp
```diff
@@ -360,26 +362,26 @@ class TmxJ2735EncodedMessage: public TmxJ2735EncodedMessageBase
 			{
 				if (msgId > MessageFrameMessage::get_default_messageId())
 				{	
-					_frame.reset(TmxJ2735EncodedMessage<MessageFrameMessage>::decode_j2735_message<
-							codec::uper<MessageFrameMessage> >(theData));
+					_frame = TmxJ2735EncodedMessage<MessageFrameMessage>::decode_j2735_message<
```
paulbourelly999, line 365: @liuji1031 Actually I think I can try to use the aliasing constructor here on line 367 for the _decoded shared pointer. Instead of passing in frame->get_j2735_data() I can pass wrap this in an aliased pointer for shared ownership. I think though I will hold off on this until this PR merges. This way I do not block this on a perfect solution

liuji1031: Sounds good. We can look into that solution in the future.

liuji1031 (approved):

src/tmx/TmxUtils/src/PluginClient.h
MishkaMN: I was wondering whether if it is possible for the user to mistake this call:
`AddMessageFilter<SpatMessage>(this, &IntersectionValidationPlugin::HandleSpatMessage);` 

with something like:

`AddMessageFilter<SpatEncodedMessage>(this, &IntersectionValidationPlugin::HandleSpatMessageEncoded);` 

where `SpatEncodedMessage = TmxJ2735EncodedMessage<SpatMessage>`

But I guess that is not possible. I think it might just fail to compile. So nevermind

PR #931, opened 2026-07-30
fix: Processing logic of raw SPDU in MessageReceiverPlugin

liuji1031:
# PR Details
## Description

Fixes secured (IEEE 1609.2) SPDU handling in the MessageReceiverPlugin. In `fullSPDUMode` the plugin was forwarding the entire SPDU — security wrapper and all — to TMX Core as the message payload, and `RawSpdu` carried only one byte field. Now the full SPDU bytes and the extracted unsecured payload are both kept in the `RawSpdu` message. If unwrapping of SPDU fails, nothing is published.

## Changes

**`RawSpdu` message**
- changed `spduData` to `fullByteData` (full SPDU bytes)
- added `msgByteData` (extracted unsecured payload).

**`MessageReceiverPlugin`**
- `fullSPDUMode` forwards the unwrapped payload to TMX Core and publishes a `RawSpdu` with both fields. Non-SPDU mode is unchanged.
- Failed or empty unwrap now short-circuits — nothing is published, `_failedSPDU` is incremented. Previously it fell through and published anyway.
- Encoding detection moved into a `get_encoding()` helper and is derived from the unsecured payload in SPDU mode, not the SPDU's first byte.
- Raised `MAX_SPDU_DEPTH` from 4 to 20; added docstrings.
- The original "Utils.h" is now moved to `tmx/Messages/include/RawSpduUtils.h"` such that other plugins can access these util functions specific to `RawSpdu` messages.
- `getJ2735SubType` method is added to `RawSpduUtils.h` which can extract message type string from the message payload. This is useful for `MessageReceiverPlugin` to assign `messageType` when constructing `RawSpdu` messages, i.e., in`buildRawSpdu` method.
- `assignMsgPSID` method is added to `RawSpduUtils.h`. It relies on `constexpr` for compile-time optimization for quick look up of message type to PSID mapping.

**`TmxApiMessages.h`**
- `MSGTYPE_PSID_PAIRS` is added to quickly look up message PSID from the message type string, such that other plugins can assign the correct PSID when routing the message for broadcasting.

**Tests**
- `RawSpduTest.cpp`: updated to test for new `fullByteData`/`msgByteData` fields.
- `RawSpduUtilsTest.cpp`: ported from `TestUtils.cpp` in the `MessageReceiverPlugin`. `testMsgTypeExtractionFromMsgBytes` is added to test `getJ2735SubType` method.

## Testing
- `RawSpduTest` and `RawSpduUtilsTest.cpp` unit tests updated and passing.

## Related Jira Key

[DT4IT-103](https://usdot-carma.atlassian.net/browse/DT4IT-103?atlOrigin=eyJpIjoiMjc0NTNiY2FjYTY1NDgxZGJlNjk5ZWJhMjY0OTU5MzMiLCJwIjoiaiJ9)

## How Has This Been Tested?

The `MessageReceiverPlugin` is built using the `build.sh` script inside the dev container and both `RawSpduTest` and `TestUtils` unit tests were confirmed to pass.

## Types of changes

## Checklist:

[DT4IT-103]: https://usdot-carma.atlassian.net/browse/DT4IT-103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/tmx/Messages/test/RawSpduTest.cpp
```diff
@@ -7,89 +7,96 @@ using namespace std;
 using namespace tmx;
 using namespace tmx::messages;
 
+// BSM Payload
 static const std::string bsmHex =
     "001425067c0eb5842562e66e8a2b9ea6c96408b97fffffff900027d9637d07d0007fff8000640fa0";
 
+// faking secured bsm raw bytes
+static const std::string rawBytesBsmHex = "3b996de867c7419ad2ec5652d133c1fb" + bsmHex + "8cfae61d57e2bfba0b68242e11d641c2";
```
paulbourelly999, line 17: This does not seem to be valid 1609.2 data. It doesn't fail your tests because we never call unwrap in it. If this is OK for you testing purposes, lets ensure to include a note that this is not valid 1609.2  data

src/tmx/Messages/test/RawSpduTest.cpp
```diff
@@ -7,89 +7,96 @@ using namespace std;
...
+
 // PSM Payload
 static const std::string psmHex =
     "00201c000002a5158048d159e14cdd338f3d4da420101effffffff00000000";
 
+// fake secured psm raw bytes
+static const std::string rawBytesPsmHex = "837c4ce3696aa2bccc4bec4e33f7c3b0" + psmHex + "b49483fd31d5dd38eb60f6f7ef93b842";
```
paulbourelly999, line 25: Same comment as above

src/tmx/Messages/test/RawSpduTest.cpp
```diff
@@ -7,89 +7,96 @@ using namespace std;
...
+    rawSpdu.set_uuid(uuid);
+    rawSpdu.set_messageType("BSM"); // BasicSafetyMessage
+    long ts = 1718900000000;
+    rawSpdu.set_timestampMs(ts);
+    rawSpdu.set_psid(8002);
 
-    EXPECT_EQ("BSM", msg.get_messageType());
-    EXPECT_EQ(1718900000000, msg.get_timestampMs());
 
+    std::cout << rawSpdu.to_string() << std::endl;
```
paulbourelly999: Remove this

src/tmx/TmxUtils/src/RawSpduUtils.h
```diff
@@ -18,99 +18,125 @@
...
+     * Returns boolean indicating success or failure. 
+     * 
+     * @param d Ieee1609Dot2Data_t pointer to the SPDU data to unwrap.
+     * @param payloadOut the byte_stream to store the extracted unsecured payload bytes.
+     * @param psidOut stores the extracted PSID if found.
+     * @param psidSet boolean flag indicating whether the PSID was set.
+     * @param depth the current recursion depth (used to prevent excessive nesting).
+     * @return bool
+     */
+    inline bool unwrapSpdu(const Ieee1609Dot2Data_t *d, std::vector<uint8_t>& payloadOut,
```
paulbourelly999, line 56: Did you change any logic in this function?

src/build.sh
```diff
@@ -19,7 +19,7 @@ set -e
 numCPU=$(nproc)
 
 show_help() {
-  echo "Usage: $0 [BUILD_TYPE] [--j2735-version <version>] [--skip-plugins 'list']"
+  echo "Usage: $0 [BUILD_TYPE] [--j2735-version <version>] [--skip-plugins 'list'] [--select-plugins 'list']"
```
paulbourelly999: What is the difference between skip plugins and select plugins. Lets pick one and stick to it. Telematics currently uses this parameter to compile their TRU so if we modify this we need to make sure they get this modification to.

paulbourelly999 (changes requested):

src/build.sh
liuji1031: The --select-plugins option is removed for now.

src/tmx/TmxUtils/src/RawSpduUtils.h
liuji1031: The logic is not changed. Only documentation blocks were added.

src/tmx/Messages/test/RawSpduTest.cpp
liuji1031: The comments above are updated to point out that the strings prepended and appended are random strings.

src/tmx/Messages/test/RawSpduTest.cpp
liuji1031: The comments are updated.

src/tmx/Messages/include/RawSpduUtils.h
```diff
@@ -0,0 +1,205 @@
...
+                }
+            }
+            else {
+                return SPDU_UNKNOWN_MESSAGE_TYPE;
+            }
+
+        }
+        catch (const std::exception &)
+        {
+            // Identifying the type is best-effort; the SPDU is still forwarded.
```
paulbourelly999: Fix the sonar code smell here. At least the one about handling the exception. Catching and exception and doing nothing is a real dangerous thing because the exception is still occurring but you are doing nothing to mitigate it or notify the user

paulbourelly999: At the very least log something here using FILE_LOG

src/tmx/TmxUtils/src/RawSpduUtils.h
paulbourelly999, line 1: Lets move this file to TmxUtils Package. I think it is more appropriate there

src/tmx/TmxApi/tmx/TmxApiMessages.h
```diff
@@ -340,6 +342,44 @@ static CONSTEXPR const char *MSGPSID_TESTMESSAGE13_PSID_STRING = "0xBFEE";
 static CONSTEXPR const char *MSGPSID_TESTMESSAGE14_PSID_STRING = "0xBFEE";
 static CONSTEXPR const char *MSGPSID_TESTMESSAGE15_PSID_STRING = "0xBFEE";
 
+// for easy lookup of PSID by message subtype string
+constexpr std::array<std::pair<std::string_view, msgPSID>, 33> MSGTYPE_PSID_PAIRS = {{
```
paulbourelly999, line 346: Why is this an array instead of a map?

src/v2i-hub/MessageReceiverPlugin/CMakeLists.txt
```diff
@@ -15,15 +15,15 @@ target_link_libraries( ${PROJECT_NAME} tmxutils ::stol-1609dot2-2022)
...
-set(BINARY ${PROJECT_NAME}_test)
-file(GLOB_RECURSE TEST_SOURCES LIST_DIRECTORIES false test/*.h test/*.cpp)
-set(SOURCES ${TEST_SOURCES} WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/test)
-add_executable(${BINARY} ${TEST_SOURCES})
-add_test(NAME ${BINARY} COMMAND ${BINARY})
-target_include_directories(${BINARY} PUBLIC  
-    src/ 
-    /opt/carma/include/stol-1609dot2-2022) 
-target_link_libraries(${BINARY} PUBLIC gtest gmock gtest_main ::stol-1609dot2-2022 tmxutils)
+# enable_testing()
```
paulbourelly999: Please delete this if its no longer necessary

paulbourelly999 (changes requested):

src/tmx/Messages/include/RawSpduUtils.h
```diff
@@ -0,0 +1,205 @@
...
+
+#include <tmx/TmxException.hpp>
+#include <tmx/messages/byte_stream.hpp>
+#include <tmx/j2735_messages/J2735MessageFactory.hpp>
+
+#include <boost/uuid/uuid.hpp>
+
+namespace tmx::messages {
+
+    static J2735MessageFactory spduUtilJ2735Factory_;
```
paulbourelly999: If you make this a static object your likely need to make it thread safe with some sort of mutex/lock. In other words, if one thread is currently using this factory, you have to prevent another thread from using it simultaneously otherwise you will have concurrency issues

src/tmx/Messages/include/RawSpduUtils.h
liuji1031: fixed: added FILE_LOG with error message.

src/tmx/TmxUtils/src/RawSpduUtils.h
liuji1031: Fixed: the RawSpduUtils.h now lives under tmx/TmxUtils/src

src/tmx/TmxApi/tmx/TmxApiMessages.h
liuji1031: The `assignMsgPSID` method is just added to the `RawSpduUtils.h`. Since the variable is constexpr the look-up uses compile time optimization to avoid runtime hashing that is otherwise needed.

src/tmx/Messages/include/RawSpduUtils.h
liuji1031: fixed: mutex guard added

src/tmx/Messages/test/RawSpduTest.cpp
liuji1031: fixed: std::cout are removed

src/v2i-hub/MessageReceiverPlugin/CMakeLists.txt
liuji1031: fixed: deleted testing section

paulbourelly999 (approved):

PR #935, opened 2026-08-05
Immediate Forward Plugin : add helper functions to improve readability and reduce nesting

liuji1031:
# PR Details
## Description

Refactors `ImmediateForwardPlugin::SendMessageToRadio` to reduce nesting and improve readability. No behavioral changes.

## Changes

- Extracted the inline HSM signing block into a new `SignWithHsm()` helper, which returns `bool` so the caller aborts the send on a signing failure (matching the previous inline `return`).
- Extracted the RSU 4.1 message formatting block into a new `ConstructMessageRSU_4_1()` helper, reformatted as a single chained `<<` expression.
- Replaced the `if (messageConfig.tmxType == msg->subtype) { ... }` wrapper with an early `continue`, removing one level of nesting from the config loops.
- Reformatted the multi-line `PLOG` statements

## Files

- `src/v2i-hub/ImmediateForwardPlugin/src/ImmediateForwardPlugin.cpp`
- `src/v2i-hub/ImmediateForwardPlugin/src/ImmediateForwardPlugin.h`

## Testing

No functional change; existing plugin behavior for RSU 4.1 (UDP) and NTCIP-1218 (SNMP) paths is unchanged, including the abort-on-signing-error path.

## Types of changes

## Checklist:

### DIFF
--- a/src/v2i-hub/ImmediateForwardPlugin/src/ImmediateForwardPlugin.cpp
+++ b/src/v2i-hub/ImmediateForwardPlugin/src/ImmediateForwardPlugin.cpp
@@ -141,6 +141,86 @@ namespace ImmediateForward
 
 	}
 
+	inline bool ImmediateForwardPlugin::SignWithHsm(const ImfConfiguration& imfConfig, const MessageConfig& messageConfig, IvpMessage* msg, string& payloadbyte)
+	{
+		std::string mType = messageConfig.sendType;
+
+		std::for_each(mType.begin(), mType.end(), [](char & c){
+			c = ::tolower(c);
+		});
+		/* convert to hex array */
+
+		string msgString=msg->payload->valuestring;
+		string base64str="";
+
+		hex2base64(msgString,base64str);
+
+		std::string req = "\'{\"type\":\""+mType+"\",\"message\":\""+base64str+"\"}\'";
+
+		string cmd1="curl -X POST " + imfConfig.hsmUrl.value() + "sign" + " -H \'Content-Type: application/json\' -d "+req;
+		const char *cmd=cmd1.c_str();
+		char buffer[2048];
+		std::string result="";
+		FILE* pipe= popen(cmd,"r");
+
+		if (pipe == NULL )
+			throw std::runtime_error("popen() failed!");
+		try{
+			while (fgets(buffer, sizeof(buffer),pipe) != NULL)
+			{
+				result+=buffer;
+			}
+		} catch (std::exception const & ex) {
+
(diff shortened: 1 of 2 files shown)


src/v2i-hub/ImmediateForwardPlugin/src/ImmediateForwardPlugin.cpp
```diff
@@ -165,121 +245,60 @@ namespace ImmediateForward
...
+				string payloadbyte="";
+
+				// Format the message using the protocol defined in the
+				// USDOT ROadside Unit Specifications Document v 4.0 Appendix C.
+
+				/// if signing is Enabled, request signing with HSM
+				if (imfConfig.enableHsm == 1)
+				{
+					bool success = SignWithHsm(imfConfig, messageConfig, msg, payloadbyte);
+					if(!success){return;}
```
paulbourelly999: Is this the logic we have initially. It suggested that when signing with HSM is unsuccessful we do nothing. Looking at the logic you moved to SignWithHsm, it looks like we include the error handling inside the method instead. I suggest we move the error handling out here so its clear what happens when this method is unsuccessful.

src/v2i-hub/ImmediateForwardPlugin/src/ImmediateForwardPlugin.h
```diff
@@ -52,6 +52,8 @@ class ImmediateForwardPlugin : public tmx::utils::PluginClient
 		void OnMessageReceived(IvpMessage *msg);
 		void OnStateChange(IvpPluginState state);
 
+		inline bool SignWithHsm(const ImfConfiguration& imfConfig, const MessageConfig& messageConfig, IvpMessage* msg, string& payloadbyte);
```
paulbourelly999: NIT: Lets write Javadoc style comments for these methods.

src/v2i-hub/ImmediateForwardPlugin/src/ImmediateForwardPlugin.cpp
liuji1031: fixed: the function now throws TmxExceptions and the caller catches them.

src/v2i-hub/ImmediateForwardPlugin/src/ImmediateForwardPlugin.h
liuji1031: fixed: comment blocks added

paulbourelly999 (approved):

PR #951, opened 2026-09-04
fix: null pointer dereference in SNMPClient::log_error

jessicaliu-oss:
## Summary
- `SNMPClient::log_error()` dereferenced `response->variables->type` whenever `status == STAT_SUCCESS`, without checking that `response` or `response->variables` were non-null — even though the same function already guards `response` correctly a few lines later before its second use. A real SNMP response can legitimately carry an empty variable-bindings list (`response->variables == nullptr`), so this was reachable in production, not just in a test.
- Also fixes the triggering test (`test_SNMPClient.cpp`): it declared a stack `snmp_pdu response;` without zero-initializing it, so `variables` held indeterminate garbage rather than a deterministic null — changed to `snmp_pdu response{};` so the test doesn't rely on undefined behavior.
- Adds 6 plugins to the README's plugin list that exist in `src/v2i-hub/` but were never documented there: `CommandPlugin`, `ERVCloudForwardingPlugin`, `IntersectionValidationPlugin`, `MUSTSensorDriverPlugin`, `PortDrayagePlugin`, `PriorityPlugin`.

## Test plan
- [x] Caught by AddressSanitizer (SEGV) while running a coverage build (`./build.sh coverage --j2735-version 2024` + `./test.sh coverage`) — crash reproduced on `develop` prior to this fix.
- [x] After the fix, `test.sh coverage` runs clean: 467 tests across 19 test binaries, 0 failures.
- [x] Verified `src/v2i-hub/` plugin directory list against the README bullet list to confirm the 6 additions are accurate and complete.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,12 @@ The V2X Hub system reduces time needed to create and deploy a roadside based V2X
 * CARMA Cloud Plugin – An interface between V2X Hub and CARMA Cloud
 * CDASim Adapter Plugin – An interface between V2X Hub and CDASim for simulation testing 
 * RSU Health Monitoring Plugin – A plugin used to monitor RSU health via RSU 4.1 specification or NTCIP 1218
+* Command Plugin – Listens for websocket connections from the TMX admin portal and processes commands from the Administration Portal
+* ERV Cloud Forwarding Plugin – Forwards BSMs broadcast by Emergency Response Vehicles (ERVs) to CARMA Cloud over a websocket connection
+* Intersection Validation Plugin – Validates MAP and SPaT messages received from RSUs against the CTI 4501 standard
+* MUST Sensor Driver Plugin – Consumes detections from an AI Waysion MUST camera sensor and translates them into Sensor Detected Object messages for cooperative perception
+* Port Drayage Plugin – Facilitates communication between infrastructure, vehicles, and container handling equipment (CHE) for automated port drayage operations
+* Priority Plugin – Handles J2735 Signal Request Messages (SRMs) and places priority requests against a Traffic Signal Controller (TSC) per NTCIP 1211
 
 V2X Hub is a communication, computation, and processing platform for V2I applications, and providing the functions listed below.
 
--- a/src/tmx/TmxUtils/src/SNMPClient.cpp
+++ b/src/tmx/TmxUtils/src/SNMPClient.cpp
@@ -391,7 +391,14 @@ namespace tmx::utils
 
         if (status == STAT_SUCCESS)
         {
-            PLOG(logERROR) << "Variable type: " << response->variables->type << ". Error in packet " << static_cast<std::string>(snmp_errstring(static_cast<int>(response->errstat)));
+            if (response && response->variables)
+            {
+                PLOG(logERROR) << "Variable type: " << response->variables->type << ". Error in packet " << static_cast<std::string>(snmp_errstring(static_cast<int>(response->errstat)));
+            }
+            else
+            {
+                PLOG(logERROR) << "Error in packet with no variables in response";
+            }
         }
         else if (status == STAT_TIMEOUT)
         {
--- a/src/tmx/TmxUtils/test/test_SNMPClient.cpp
+++ b/src/tmx/TmxUtils/test/test_SNMPClient.cpp
@@ -62,7 +62,7 @@ namespace unit_test
 
     TEST_F(test_SNMPClient, log_error)
     {
-        snmp_pdu response;
+        snmp_pdu response{};
         ASSERT_NO_THROW(scPtr->log_error(STAT_ERROR, request_type::GET, &response));
         ASSERT_NO_THROW(scPtr->log_error(STAT_ERROR, request_type::SET, &response));
         ASSERT_NO_THROW(scPtr->log_error(STAT_SUCCESS, request_type::OTHER, &response));


paulbourelly999:
The test plan section of this indicates that there is a failure in the unit tests on develop. We are not experiencing any failure or AddressSanitizer (SEGV) error in our CI builds or on local testing on develop branch. Please describe the crash or issue you were running into on develop before hand.

jessicaliu-oss:
- Hello Paul,

Below information confirms how the issue was spotted. I ran both coverage
and Valgrind on the recently downloaded code, and both tools identified
this issue. I consider this a solid issue, so go ahead and check in the
fixes. If this causes any problems, I apologize. I also found other minor
bugs via these two tools while setting up my build routine. Because of this
same concern, I kept those fixes in my local branch. I checked in too
quickly before fully understanding the process.

*How *to find*:* check out the exact pre-fix commit (f5141471, parent of
the fix), build tmx/TmxUtils with -fsanitize=address, run the test binary.
- *The test case:* test_SNMPClient.cpp:68 — TEST_F(test_SNMPClient,
log_error), the third assertion, scPtr->log_error(STAT_SUCCESS,
request_type::OTHER, &response).
- *The crash site:* SNMPClient.cpp:394 — exactly the line the diff touched (
response->variables->type).
- *Address 0x18*, not 0x0 — ASan calls out "address points to the zero
page," i.e. this is a near-null offset, consistent with a small/garbage
pointer value in response.variables (offset 0x18 is variables' byte offset
within a mostly-zeroed-by-luck-but-not-guaranteed stack frame, not a
deliberately poisoned address) — matching "indeterminate stack garbage that
happened to be small/invalid," not a deterministic fixed bug pattern.

AddressSanitizer:DEADLYSIGNAL
==1840318==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000018
==1840318==The signal is caused by a READ memory access.
==1840318==Hint: address points to the zero page.
    #0 ... tmx::utils::snmp_client::log_error(...) SNMPClient.cpp:394
    #1 ... test_SNMPClient_log_error_Test::TestBody() test_SNMPClient.cpp:68
SUMMARY: AddressSanitizer: SEGV .../SNMPClient.cpp:394 in ...log_error(...)

Separately, even in the abstract, the C++ standard version wouldn't matter
here. The relevant rule is:

   - snmp_pdu response; (no braces) → *default-initialization*. Since
   snmp_pdu has a trivial default constructor (it's a plain C struct —
   scalars and pointers, no user-defined constructor), default-initialization
   performs *no initialization at all* — every field, including variables,
   holds indeterminate garbage.
   - snmp_pdu response{}; (the fix) → *value-initialization*, which
   zero-initializes a trivial type before running its (trivial, no-op) default
   constructor.

paulbourelly999:
Thanks for you response and contribution. I have done my own investigation and have found the following.
I have confirmed that the ASAN error does not appear in develop as is and this is partially due to the bug in the unit test you are fixing here.

Specifically when initializing the snmp_pdu without zeroing it's fields like this
```cpp
 TEST_F(test_SNMPClient, log_error)
    {
        snmp_pdu response;
        ASSERT_NO_THROW(scPtr->log_error(STAT_ERROR, request_type::GET, &response));
        ASSERT_NO_THROW(scPtr->log_error(STAT_ERROR, request_type::SET, &response));
        ASSERT_NO_THROW(scPtr->log_error(STAT_SUCCESS, request_type::OTHER, &response));
        ASSERT_NO_THROW(scPtr->log_error(STAT_TIMEOUT, request_type::OTHER, &response));
    }
```
calls to log_error are passed a garbage initialized response object, which is not null. I think ASan is fine with this since it seems like a regular non null object.

When adding the zero initialization however ASan does though a SEGV
```cpp
 TEST_F(test_SNMPClient, log_error)
    {
        snmp_pdu response{}; // zero/null initialization of member variables 
        ASSERT_NO_THROW(scPtr->log_error(STAT_ERROR, request_type::GET, &response));
        ASSERT_NO_THROW(scPtr->log_error(STAT_ERROR, request_type::SET, &response));
        ASSERT_NO_THROW(scPtr->log_error(STAT_SUCCESS, request_type::OTHER, &response));
        ASSERT_NO_THROW(scPtr->log_error(STAT_TIMEOUT, request_type::OTHER, &response));
    }
```

```
==27626==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000018 (pc 0x57de4da8437b bp 0x7ffe775b7820 sp 0x7ffe775b6d70 T0)
==27626==The signal is caused by a READ memory access.
==27626==Hint: address points to the zero page.
    #0 0x57de4da8437b in tmx::utils::snmp_client::log_error(int const&, tmx::utils::request_type const&, snmp_pdu const*) const /workspace/src/tmx/TmxUtils/src/SNMPClient.cpp:394
    #1 0x57de4d9fbae6 in unit_test::test_SNMPClient_log_error_Test::TestBody() /workspace/src/tmx/TmxUtils/test/test_SNMPClient.cpp:68
    #2 0x57de4db2405e in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (/workspace/src/build/tmx/TmxUtils/tmxutils_test+0xc5405e)
    #3 0x57de4db192e5 in testing::Test::Run() (/workspace/src/build/tmx/TmxUtils/tmxutils_test+0xc492e5)
```

I think this is because we are attempting to read pointer members from this members that are null initialized like here where we are reading `response->variables->type` where `variables=nullptr`

```cpp
void snmp_client::log_error(const int &status, const request_type &request_type, const snmp_pdu *response) const
    {

        if (status == STAT_SUCCESS)
        {
            PLOG(logERROR) << "Variable type: " << response->variables->type << ". Error in packet " << static_cast<std::string>(snmp_errstring(static_cast<int>(response->errstat)));
        }
``` 

Your fix here only works because you also check whether variables is not null. Without this check the ASan error will still be present since response is not null but variable is.

The larger question I think here is which of these represent values that the net-snmp library will return. Meaning does net-snmp ever return a non-null response with a null variable type when encountering an error. 

From a brief look at the net_snmp implementation of snmp_synch_response it looks like in any error case the entire response pdu will be null which does mean our log_error function should protect against that. I think the correct fix here is 

```cpp
  TEST_F(test_SNMPClient, log_error)
    {
        //snmp_pdu response will be nullptr on error from snmp_synch_response
        EXPECT_NO_THROW(scPtr->log_error(STAT_ERROR, request_type::GET, nullptr));
        EXPECT_NO_THROW(scPtr->log_error(STAT_ERROR, request_type::SET, nullptr));
        EXPECT_NO_THROW(scPtr->log_error(STAT_SUCCESS, request_type::OTHER, nullptr));
        EXPECT_NO_THROW(scPtr->log_error(STAT_TIMEOUT, request_type::OTHER, nullptr));
    }
```
and 

```
void snmp_client::log_error(const int &status, const request_type &request_type, const snmp_pdu *response) const
    {

        if (status == STAT_SUCCESS && response)
        {
            PLOG(logERROR) << "Variable type: " << response->variables->type << ". Error in packet " << static_cast<std::string>(snmp_errstring(static_cast<int>(response->errstat)));
        }
        else if (status == STAT_TIMEOUT)
        {
            PLOG(logERROR) << "Timeout, no response from server";
        }
        else
        {
            PLOG(logERROR) << "Unknown SNMP Error (status code: " << status << ") for " << (request_type == request_type::GET ? "GET" : "SET");
        }

        if (response)
        {
            PLOG(logERROR) << "SNMP error in response: " << snmp_errstring(response->errstat);
        }

    }
```

Does this satisfy your testing requirements? Please let me know if you think I have made and error somewhere here.

jessicaliu-oss:
SNMP error in response would never print since response is a nulltpr. Should it be if(!response) instead?


  TEST_F(test_SNMPClient, log_error)
    {
        //snmp_pdu response will be nullptr on error from snmp_synch_response
        EXPECT_NO_THROW(scPtr->log_error(STAT_ERROR, request_type::GET, nullptr));
        EXPECT_NO_THROW(scPtr->log_error(STAT_ERROR, request_type::SET, nullptr));
        EXPECT_NO_THROW(scPtr->log_error(STAT_SUCCESS, request_type::OTHER, nullptr));
        EXPECT_NO_THROW(scPtr->log_error(STAT_TIMEOUT, request_type::OTHER, nullptr));
    }


void snmp_client::log_error(const int &status, const request_type &request_type, const snmp_pdu *response) const

{
...

        if (response)
        {
            PLOG(logERROR) << "SNMP error in response: " << snmp_errstring(response->errstat);
        }

}

It will be great to run via Valgrind as a dynamic memory check tool. Valgrind is much through and captures real bugs.



________________________________

paulbourelly999:
There are a couple parts to this. The first is whether this is a scenario that can occur while running the software. Meaning is there a case in which `snmp_sync_response` will return a non null snmp_pdu when encountering an error? For this I am unsure but having looked at the code briefly I did not see a case since the response snmp_pdu looks like it would be set to null in any error case (https://github.com/haad/net-snmp/blob/master/snmplib/snmp_client.c).

The second part is whether there is unit test code coverage for this entire log_error function. For that I think we could just write another unit test that was something like 
```cpp
  TEST_F(test_SNMPClient, log_error_non_null_response)
    {
        snmp_pdu response{};  
        response.variables = somelistofvariables;
        response.errstat = someerrorcode;

        EXPECT_NO_THROW(scPtr->log_error(STAT_ERROR, request_type::GET, &response));
        EXPECT_NO_THROW(scPtr->log_error(STAT_ERROR, request_type::SET, &response));
        EXPECT_NO_THROW(scPtr->log_error(STAT_SUCCESS, request_type::OTHER, &response));
        EXPECT_NO_THROW(scPtr->log_error(STAT_TIMEOUT, request_type::OTHER, &response));
    }
```

This would then cover the code in question with a unit test, but I would argue, there is very little value in covering this code with a unit test if :

- We are unsure it is an actual case that can be encountered
- We are unsure our unit test represents an actual state that could be encountered.

Lastly you mentioned running our code through Valgrind and finding bugs. While we have done this for some of our code we are not able to do it for all the code we develop and currently rely mostly on Asan and Sonar to find these types of bugs.

Valgrind can detect some bugs especially with memory management both of these tools miss. Please let us know if you have discovered any such issues. We are very interested in fixing these

jessicaliu-oss:
The fixes I checked in are verified by both coverage and Valgrind with 100% pass. I agree that may not happen. From the theoretical prospect, this can be fixed simply. I would leave it to you since I don't run ASAN but am aware it has limitation. Enjoy long weekend.
________________________________

jessicaliu-oss:
The fixes I checked in are verified by both coverage and Valgrind with 100% pass. I agree that may not happen. From the theoretical prospect, this can be fixed simply. I would leave it to you since I don't run ASAN but am aware it has limitation.

Enjoy memorial long weekend.

________________________________

PR #952, opened 2026-09-04
fix: use-after-free issue when consuming kafka payloads from sensor_data_sharing_service

liuji1031:
# PR Details
## Description

This PR fixes the following issue:

`CARMAStreetsPlugin` fails to parse SDSM JSON consumed from the `v2xhub_sdsm_sub` topic. Every failure truncates the payload at 688 bytes, with the JSON parser reporting `Missing ',' or '}'` at column 687 and a trailing garbage byte (`P\x03`) where JSON should continue.

An example console log is as follows:
```bash
[2026-09-04 18:43:47.381] /kafka_consumer_worker.cpp (153) - DEBUG1 : rdkafka#consumer-5 message Consumed: 691 bytes : {"msg_cnt":122,"source_id":"rsu_1234","equipment_type":1,"sdsm_time_stamp":{"second":47374,"minute":43,"hour":18,"day":4,"month":9,"year":2026,"offset":0},"ref_pos":{"long":-771483512,"lat":389519791,"elevation":0},"ref_pos_xy_conf":{"semi_major":0,"semi_minor":0,"orientation":0},"objects":[{"detected_object_data":{"detected_object_common_data":{"obj_type":2,"object_id":13684,"obj_type_cfd":100,"measurement_time":87,"time_confidence":0,"pos":{"offset_x":3254,"offset_y":-99,"offset_z":0},"pos_confidence":{"pos":15,"elevation":15},"speed":0,"speed_confidence":7,"speed_z":0,"speed_confidence_z":4,"heading":0,"heading_conf":4},"detected_object_optional_data":{"detected_vru_data":{}}}}]}

[...]

PR #953, opened 2026-09-08
[DRAFT] test: cover consuming a kafka payload that is not null terminated

MishkaMN:
# PR Details
## Description

Adds the regression test for the payload truncation fixed in #952. Targets that PR's branch so it merges together with the fix.

The payload is 15 bytes with a NUL at index 7: `{"a":1}\0{"b":2}`. Read as a C string it is only 7 bytes long, so the test asserts both the length and the full content of what `consume()` returns.

It uses librdkafka's in-process mock cluster (`rd_kafka_mock_cluster_new`), so:
- no broker and no CI change are needed, and
- only the public `consume()` API is exercised, so `msg_consume` stays private.

## Related GitHub Issue

N/A

## Related Jira Key

N/A

## Motivation and Context

The existing `test_kafka_consumer_worker` runs with no broker, so `consume()` returns on the `ERR__TIMED_OUT` branch. The bug lived in the `ERR_NO_ERROR` branch, which no test reached — that test passes identically before and after the fix.

## How Has This Been Tested?

Compiled and run against both versions of `msg_consume` in the V2X-Hub image:

| `msg_consume` | produced | consumed | equal |
|---|---|---|---|
| with the fix (`assign(payload, len())`) | 15 | 15 | yes |
| before the fix (`= static_cast<const char*>(payload())`) | 15 | 7 | no |

Note on timing: `kafka_consumer_worker::init()` does not set `auto.offset.reset`, so the consumer starts at the latest offset. A single produce before subscribing is missed, so the test produces inside the poll loop until the consumer has joined the group and been assigned the partition.

## Types of changes

## Checklist:

### DIFF
--- a/src/tmx/TmxUtils/src/kafka/kafka_consumer_worker.cpp
+++ b/src/tmx/TmxUtils/src/kafka/kafka_consumer_worker.cpp
@@ -122,11 +122,10 @@ namespace tmx::utils
         }
     }
 
-    const char *kafka_consumer_worker::consume(int timeout_ms)
+    std::string kafka_consumer_worker::consume(int timeout_ms)
     {
         std::shared_ptr<RdKafka::Message> msg(_consumer->consume(timeout_ms));
-        const char *msg_str = msg_consume(msg);
-        return msg_str;
+        return msg_consume(msg);
     }
 
     bool kafka_consumer_worker::is_running() const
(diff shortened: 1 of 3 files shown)


MishkaMN:
Running CI against develop on https://github.com/usdot-fhwa-OPS/V2X-Hub/actions/runs/34220260222

